PR #115, opened 2026-01-15
NOISSUE - Deploy proplet on kata containers

SammyOina:
# What type of PR is this?

### DIFF
--- a/coco/deploy_coco.sh
+++ b/coco/deploy_coco.sh
@@ -11,6 +11,7 @@ K8S_DIR="$SCRIPT_DIR"
 # Configuration
 IMAGE_NAME="ghcr.io/absmach/propeller/proplet"
 IMAGE_TAG="latest"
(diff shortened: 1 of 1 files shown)


coco/deploy_coco.sh
```diff
@@ -0,0 +1,43 @@
...
+set -e
+
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
+ROOT_DIR=$(dirname "$SCRIPT_DIR")
+K8S_DIR="$SCRIPT_DIR"
+
+# Configuration
+IMAGE_NAME="ghcr.io/absmach/propeller/proplet"
+IMAGE_TAG="latest"
+RUNTIME_CLASS=${RUNTIME_CLASS:-kata}
```
danko-miladinovic: Can you please add a comment that states that the user should define the variable `RUNTIME_CLASS` with the runtime class that he wants to use with CoCo.

PR #133, opened 2026-02-05
NOISSUE - Add scheduled cron tasks and task priority scheduling

JeffMboya:
# What type of PR is this?
Feature

## What does this do?

Adds scheduled cron tasks and task priority scheduling

No issue

### DIFF
--- a/pkg/cron/parser.go
+++ b/pkg/cron/parser.go
@@ -11,7 +11,7 @@ var ErrInvalidCronExpression = errors.New("invalid cron expression")
 
 // Schedule represents a parsed cron schedule that can compute the next run time.
 type Schedule interface {
-	Next(time.Time) time.Time
+	Next(t time.Time) time.Time
 }
 
 type cronSchedule struct {
--- a/manager/service.go
+++ b/manager/service.go
@@ -781,6 +781,14 @@ func (svc *service) GetParentResults(ctx context.Context, taskID string) (map[st
 	return parentResults, nil
 }
 
+func (svc *service) StartCronScheduler(ctx context.Context) error {
+	if svc.cronScheduler == nil {
+		return nil
+	}
+
+	return svc.cronScheduler.Start(ctx)
+}
+
 func (svc *service) checkTaskDependencies(ctx context.Context, taskID string, t task.Task) error {
 	allDepsCompleted := true
 	for _, depID := range t.DependsOn {
@@ -948,14 +956,6 @@ func (svc *service) stopJobTasks(ctx context.Context, tasks []task.Task) {
 			_ = svc.StopTask(ctx, t.ID)
 		}
 	}
-}
-
-func (svc *service) StartCronScheduler(ctx context.Context) error {
-	if svc.cronScheduler == nil {
-		return nil
-	}
-
-	return svc.cronScheduler.Start(ctx)
 }
 
 func (svc *service) handle(ctx context.Context) func(topic string, msg map[string]any) error {


JeffMboya:
> I think this should wait for storage so that we can store the crons on persistent storage

Implemented file-based persistent storage.

pkg/storage/file.go
```diff
@@ -0,0 +1,300 @@
+package storage
```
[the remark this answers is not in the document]
JeffMboya: > Let's not use file-based storage, we will use PostgreSQL, SQLite (file) or Badger

I have replaced file storage with badger. 

## How it works
- When you create a task with a `Schedule` field, it's stored in Badger via `tasksDB.Create()`
- The cron scheduler reads from Badger every minute to find tasks ready to run
- Task schedules, `NextRun` times, and recurring flags are persisted in Badger
- On restart, the cron scheduler loads all scheduled tasks from Badger and recalculates `NextRun` times if needed

Badger is the default storage backend, so scheduled tasks are automatically persisted in Badger unless you explicitly set `MANAGER_STORAGE_TYPE=memory`.

pkg/storage/badger.go
```diff
@@ -0,0 +1,322 @@
...
+	Value json.RawMessage `json:"value"`
+}
+
+type badgerStorage struct {
+	sync.RWMutex
+
+	db *badger.DB
+}
+
+func NewBadgerStorage(dataDir string) (Storage, error) {
```
rodneyosodo: Why not use pkg/storage/badger ?

cmd/manager/main.go
```diff
@@ -117,7 +116,14 @@ func main() {
 		return
 	}
 
-	repos, err := storage.NewRepositories(cfg.Storage)
+	storageCfg := storage.Config{}
+	if err := env.Parse(&storageCfg); err != nil {
```
rodneyosodo: Should we add env prefix option?

pkg/storage/factory.go
```diff
@@ -24,6 +24,8 @@ type Config struct {
 	SQLitePath string `env:"MANAGER_SQLITE_PATH" envDefault:"./propeller.db"`
 
 	BadgerPath string `env:"MANAGER_BADGER_PATH" envDefault:"./data/badger"`
+
+	DataDir string `env:"MANAGER_DATA_DIR" envDefault:"./data"`
```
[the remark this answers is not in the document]
JeffMboya: SQLitePath used for SQLite connection. BadgerPath used for Badger database directory

cmd/manager/main.go
JeffMboya: No, we don't need to add an env prefix option here because `storage.Config` already includes the full env var names in the tags i.e

```bash
env:"MANAGER_STORAGE_TYPE"
env:"MANAGER_POSTGRES_HOST"
env:"MANAGER_SQLITE_PATH
```

manager/cron.go
```diff
@@ -0,0 +1,218 @@
...
+	"log/slog"
+	"time"
+
+	"github.com/absmach/propeller/pkg/cron"
+	"github.com/absmach/propeller/pkg/storage"
+	"github.com/absmach/propeller/task"
+)
+
+const (
+	defaultCronCheckInterval = 1 * time.Minute
```
rodneyosodo:
```suggestion
const defaultCronCheckInterval = 1 * time.Minute
```

pkg/scheduler/priority.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"sort"
+
+	"github.com/absmach/propeller/pkg/proplet"
+	"github.com/absmach/propeller/task"
+)
+
+const defaultPriority = 50
+
+type priorityScheduler struct {
```
rodneyosodo: If `priorityScheduler` implements `Scheduler` interface, why do we have it import a scheduler?

rodneyosodo (changes requested):
Can you also update the API docs with the task config for using this

pkg/cron/parser.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var ErrInvalidCronExpression = errors.New("invalid cron expression")
+
+type CronSchedule struct {
+	parser cron.Parser
+	spec   cron.Schedule
+}
+
+func ParseCronExpression(expr string) (*CronSchedule, error) {
```
rodneyosodo: Same

PR #139, opened 2026-02-11
PROP-129 - Graceful shutdown and restart

### LINKED ISSUE #129: Feature: Graceful Shutdown and Restart

When the manager or proplet services are terminated (via SIGTERM, SIGINT, or deployment updates), running tasks are abruptly stopped without cleanup. This causes:
  - Task state becomes inconsistent (tasks marked "Running" but not actually executing)
  - Resources not released (memory leaks, open connections)
  - No way to resume interrupted workloads
  - Lost task results and partial execution
 
Currently, services just exit immediately, leaving zombie task states in the in-memory storage.

Implement graceful shutdown that:
  1. **Catches termination signals** (SIGTERM, SIGINT) in both manager and proplet services
  2. **Stops all running tasks cleanly** before shutting down
  3. **Updates task states** to reflect the interruption (e.g., new "Interrupted" state)
  4. **Releases resources** (closes connections, stops monitoring, cleans up temporary files)
  5. **Supports task resumption** by saving checkpoint data for later restart

Use cases:
  - Deploying updates to manager/proplet without disrupting ongoing workloads
  - Scaling down proplet nodes during load reduction
  - Recovering from service crashes with minimal data loss
  - Rolling updates in production environments

Must-have

### Anything else?

_No response_

JeffMboya:
# What type of PR is this?
Feature

## What does this do?
Adds graceful shutdown/restart

- - Resolves #129 

Yes

No

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -30,10 +30,10 @@ import (
 )
 
 const (
-	svcName        = "manager"
-	defHTTPPort    = "7070"
-	envPrefixHTTP  = "MANAGER_HTTP_"
-	configPath     = "config.toml"
+	svcName         = "manager"
+	defHTTPPort     = "7070"
+	envPrefixHTTP   = "MANAGER_HTTP_"
+	configPath      = "config.toml"
 	shutdownTimeout = 30 * time.Second
 )
 


pkg/storage/factory.go
```diff
@@ -33,7 +34,7 @@ type Repositories struct {
 	Metrics      MetricsRepository
 }
 
-func NewRepositories(cfg Config) (*Repositories, error) {
+func NewRepositories(cfg Config) (*Repositories, io.Closer, error) {
```
rodneyosodo: Why do we need to return `io.Closer` ?

manager/workflow.go
```diff
@@ -135,8 +135,28 @@ func (wc *WorkflowCoordinator) OnTaskCompletion(ctx context.Context, taskID stri
 	return wc.CheckAndStartReadyTasks(ctx, t.WorkflowID)
 }
 
+func (wc *WorkflowCoordinator) listAllTasks(ctx context.Context) ([]task.Task, error) {
+	const pageSize uint64 = 500
```
rodneyosodo: Let's use 100

manager/workflow.go
```diff
@@ -135,8 +135,28 @@ func (wc *WorkflowCoordinator) OnTaskCompletion(ctx context.Context, taskID stri
...
+			break
+		}
+	}
+
+	return allTasks, nil
+}
+
 func (wc *WorkflowCoordinator) getWorkflowTasks(ctx context.Context, workflowID string) ([]task.Task, error) {
-	allTasks, _, err := wc.taskRepo.List(ctx, 0, 10000)
+	allTasks, err := wc.listAllTasks(ctx)
```
rodneyosodo: Should we filter by workflow id rather than returning all tasks, then filtering?

manager/service.go
```diff
@@ -789,14 +803,93 @@ func (svc *service) StartCronScheduler(ctx context.Context) error {
 	return svc.cronScheduler.Start(ctx)
 }
 
+func (svc *service) Shutdown(ctx context.Context) error {
```
rodneyosodo: Should we start the tasks after a previous shutdown?

For example, I have a task that takes 10 minutes, and then there is a shutdown. Should we try to start the task after we are back up so it completes?

rodneyosodo (changes requested):

pkg/storage/factory.go
JeffMboya: > Why do we need to return `io.Closer` ?

`NewRepositories` can initialize persistent storage backends (`postgres`, `sqlite`, `badger`) that own long-lived DB resources, so we need to return an `io.Closer` so the `manager` can release them on shutdown

manager/service.go
JeffMboya: We should not auto-start interrupted tasks by default. After restart, execution status is uncertain and auto-restart risks duplicate side effects. Safer default is mark interrupted as failed and retry explicitly; auto-restart should be opt-in for idempotent/resumable tasks

cmd/manager/main.go
```diff
@@ -160,17 +175,32 @@ func main() {
...
+		return server.StopSignalHandler(ctx, stop, logger, svcName, hs)
 	})
 
 	if err := g.Wait(); err != nil {
 		logger.Error(fmt.Sprintf("%s service exited with error: %s", svcName, err))
 	}
+
+	// Coordinated shutdown.
+	logger.Info("initiating graceful shutdown")
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), shutdownTimout)
```
rodneyosodo:
```suggestion
	shutdownCtx, shutdownCancel := context.WithTimeout(ctx, shutdownTimeout)
```

rodneyosodo (changes requested):

pkg/storage/memory_adapter.go
```diff
@@ -59,6 +59,36 @@ func (r *memoryTaskRepo) List(ctx context.Context, offset, limit uint64) ([]task
 	return tasks, total, nil
 }
 
+func (r *memoryTaskRepo) ListByWorkflowID(ctx context.Context, workflowID string, offset, limit uint64) ([]task.Task, uint64, error) {
+	// Fetch all items from the in-memory store, then filter by workflowID.
+	// The underlying storage is fully in-process, so this is fast and avoids
+	// duplicating index infrastructure in the generic key-value store.
+	const maxFetch = 100000
```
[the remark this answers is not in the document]
JeffMboya: > Why 100K, why not iteratively filtering?

Addressed in https://github.com/absmach/propeller/pull/145

PR #169, opened 2026-03-23
NOISSUE - Add wasm http url

JeffMboya:
# What type of PR is this?

This is a feature - adds wasm HTTP URL delivery

## What does this do?

Proplets can now fetch wasm binaries via plain `HTTP URL`. Adds `wasm_http_url` field to tasks, with a 100MB streaming size limit and URL scheme validation

- No related issue

Yes - unit tests for fetch logic and SQLite round-trips

Documented at https://github.com/absmach/propeller-website/pull/30


rodneyosodo (changes requested):

rodneyosodo (commented):
### Off-by-one in streaming chunk check (proplet/src/service.rs)

The bounds check at line 1280 happens *after* extending the buffer. This causes binaries exactly at the 100MB limit to be incorrectly rejected. Should check before extending:

```rust
// Current: extends then checks (incorrect)
binary.extend_from_slice(&chunk);
if binary.len() > WASM_FETCH_MAX_BYTES { // Off by one

// Should be: check before extending
if binary.len() + chunk.len() > WASM_FETCH_MAX_BYTES {
```

Affects edge case at exactly 100MB.

### Missing Content-Type validation (proplet/src/service.rs)

The `fetch_wasm_from_http` function doesn't validate the response `Content-Type` header. A server returning HTTP 200 with an error page (e.g., HTML error) would still be treated as valid WASM. Consider checking for `application/wasm` or `application/octet-stream`.

### Test coverage suggestion

Consider adding a test for streaming requests without `Content-Length` header (chunked transfer encoding) to ensure the streaming path is fully covered.

JeffMboya:
Addressed all three points from the review. 
- The off-by-one was fixed by checking `binary.len() + chunk.len() > WASM_FETCH_MAX_BYTES` before extending the buffer, so payloads at exactly 100 MB are no longer incorrectly rejected. 
- Content-Type validation was added after the status check: if the header is present and is neither `application/wasm` nor `application/octet-stream`, the fetch is rejected with a clear error - absent headers are still allowed for servers that omit the type. 
- Two tests were added: one covering HTML error pages returned with HTTP 200, and one covering streaming accumulation with no `Content-Length` header.

proplet/src/service.rs
```diff
@@ -1,11 +1,14 @@
 use crate::config::PropletConfig;
+
+const WASM_FETCH_MAX_BYTES: usize = 100 * 1024 * 1024;
```
rodneyosodo:
```suggestion
const WASM_FETCH_MAX_BYTES: usize = 100 * 1024 * 1024; /// 100MB
```

proplet/src/service.rs
```diff
@@ -1218,3 +1237,193 @@ fn build_fl_update_envelope(
...
+            return Err(anyhow::anyhow!(
+                "wasm response from {} exceeds size limit ({} bytes)",
+                url,
+                WASM_FETCH_MAX_BYTES
+            ));
+        }
+        binary.extend_from_slice(&chunk);
+    }
+
+    info!("Fetched wasm from {}, size: {} bytes", url, binary.len());
```
rodneyosodo: This should be debug

rodneyosodo (changes requested):

PR #177, opened 2026-03-30
NOISSUE - Add task metadata

JeffMboya:
# What type of PR is this?

This is a feature that adds an optional `metadata` field to tasks.

## What does this do?

Adds a `metadata map[string]any` field to the `Task` struct. Metadata is persisted as JSON in Postgres, SQLite, and Badger.

Validates metadata on create/update: must be valid JSON and under 64KB.

Adds `ListByMetadataFilter` to the repository interface, letting callers filter tasks by key-value metadata pairs.

Adds `ListTasksByFilter` to the service and exposes it via the existing list-tasks HTTP endpoint using a new `listTasksReq` type.

Simplifies `publishStart` to publish the full task struct instead of a hand-built map.

N/A

Yes, mocks and storage adapters are updated with the new method.

Documented at https://github.com/absmach/propeller-website/pull/30

### DIFF
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -263,35 +263,25 @@ func stopJobEndpoint(svc manager.Service) endpoint.Endpoint {
 
 func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request any) (any, error) {
-		switch req := request.(type) {
-		case listEntityReq:
-			if err := req.validate(); err != nil {
-				return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
-			}
-			tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
-			if err != nil {
-				return listTaskResponse{}, err
-			}
-
-			return listTaskResponse{TaskPage: tasks}, nil
-		case listTasksReq:
-			if err := req.validate(); err != nil {
-				return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
-			}
-			pm := sdk.PageMetadata{
-				Offset:         req.offset,
-				Limit:          req.limit,
-				MetadataFilter: req.metadataFilter,
-			}
-			tasks, err := svc.ListTasksByFilter(ctx, pm)
-			if err != nil {
-				return listTaskResponse{}, err
-			}
-
-			return listTaskResponse{TaskPage: tasks}, nil
-		default:
+		req, ok := request.(listTasksReq)
+		if !ok {
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, pkgerrors.ErrInvalidData)
 		}
+		if err := req.validate(); err != nil {
+			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
+		}
+
+		pm := sdk.PageMetadata{
+			Offset:         req.offset,
+			Limit:          req.limit,
+			MetadataFilter: req.metadataFilter,
+		}
+		tasks, err := svc.ListTasks(ctx, pm)
+		if err != nil {
+			return listTaskResponse{}, err
+		}
+
+		return listTaskResponse{TaskPage: tasks}, nil
 	}
 }
 
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -7,6 +7,7 @@ import (
 	apiutil "github.com/absmach/magistrala/api/http/util"
 	"github.com/absmach/propeller/manager"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
+	"github.com/absmach/propeller/pkg/sdk"
 	"github.com/go-kit/kit/endpoint"
 )
 
@@ -262,22 +263,35 @@ func stopJobEndpoint(svc manager.Service) endpoint.Endpoint {
 
 func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request any) (any, error) {
-		req, ok := request.(listEntityReq)
-		if !ok {
+		switch req := request.(type) {
+		case listEntityReq:
+			if err := req.validate(); err != nil {
+				return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
+			}
+			tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
+			if err != nil {
+				return listTaskResponse{}, err
+			}
+
+			return listTaskResponse{TaskPage: tasks}, nil
+		case listTasksReq:
+			if err := req.validate(); err != nil {
+				return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
+			}
+			pm := sdk.PageMetadata{
+				Offset:         req.offset,
+				Limit:          req.limit,
+				MetadataFilter: req.metadataFilter,
+			}
+			tasks, err := svc.ListTasksByFilter(ctx, pm)
+			if err != nil {
+				return listTaskResponse{}, err
+			}
+
+			return listTaskResponse{TaskPage: tasks}, nil
+		default:
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, pkgerrors.ErrInvalidData)
 		}
-		if err := req.validate(); err != nil {
-			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
-		}
-
-		tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
-		if err != nil {
-			return listTaskResponse{}, err
-		}
-
-		return listTaskResponse{
-			TaskPage: tasks,
-		}, nil
 	}
 }
 
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -272,9 +272,9 @@ func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 		}
 
 		pm := sdk.PageMetadata{
-			Offset:         req.offset,
-			Limit:          req.limit,
-			MetadataFilter: req.metadataFilter,
+			Offset:   req.offset,
+			Limit:    req.limit,
+			Metadata: req.metadata,
 		}
 		tasks, err := svc.ListTasks(ctx, pm)
 		if err != nil {
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -7,7 +7,6 @@ import (
 	apiutil "github.com/absmach/magistrala/api/http/util"
 	"github.com/absmach/propeller/manager"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
-	"github.com/absmach/propeller/pkg/sdk"
 	"github.com/go-kit/kit/endpoint"
 )
 
@@ -271,7 +270,7 @@ func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
 
-		pm := sdk.PageMetadata{
+		pm := manager.PageMetadata{
 			Offset:   req.offset,
 			Limit:    req.limit,
 			Metadata: req.metadata,
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -15,7 +15,10 @@ import (
 
 var errStatusFilterUnsupported = errors.New("status filter is not supported")
 
-const maxMetadataBytes = 1048576
+const (
+	maxMetadataBytes = 1048576 // 1MB
+	maxLimitSize     = 100
+)
 
 type taskReq struct {
 	task.Task `json:",inline"`
@@ -161,6 +164,10 @@ type listTasksReq struct {
 }
 
 func (r *listTasksReq) validate() error {
+	if r.limit > maxLimitSize || r.limit < 1 {
+		return apiutil.ErrLimitSize
+	}
+
 	return nil
 }
 
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -161,7 +161,7 @@ type listTasksReq struct {
 	metadata task.Metadata
 }
 
-func (r *listTasksReq) validate() error {
+func (r listTasksReq) validate() error {
 	if r.limit > api.MaxLimitSize || r.limit < 1 {
 		return apiutil.ErrLimitSize
 	}
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -158,7 +158,7 @@ func (e listEntityReq) validate() error {
 type listTasksReq struct {
 	offset   uint64
 	limit    uint64
-	metadata map[string]any
+	metadata task.Metadata
 }
 
 func (r *listTasksReq) validate() error {
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -15,7 +15,7 @@ import (
 
 var errStatusFilterUnsupported = errors.New("status filter is not supported")
 
-const maxMetadataBytes = 65536
+const maxMetadataBytes = 1048576
 
 type taskReq struct {
 	task.Task `json:",inline"`
@@ -50,7 +50,7 @@ func (t *taskReq) validate() error {
 			return fmt.Errorf("invalid metadata: %w", err)
 		}
 		if len(b) > maxMetadataBytes {
-			return errors.New("metadata exceeds 64KB limit")
+			return errors.New("metadata exceeds 1MB limit")
 		}
 	}
 
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	apiutil "github.com/absmach/magistrala/api/http/util"
+	"github.com/absmach/propeller/pkg/api"
 	"github.com/absmach/propeller/pkg/cron"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
 	"github.com/absmach/propeller/pkg/proplet"
@@ -15,10 +16,7 @@ import (
 
 var errStatusFilterUnsupported = errors.New("status filter is not supported")
 
-const (
-	maxMetadataBytes = 1048576 // 1MB
-	maxLimitSize     = 100
-)
+const maxMetadataBytes = 1048576 // 1MB
 
 type taskReq struct {
 	task.Task `json:",inline"`
@@ -164,7 +162,7 @@ type listTasksReq struct {
 }
 
 func (r *listTasksReq) validate() error {
-	if r.limit > maxLimitSize || r.limit < 1 {
+	if r.limit > api.MaxLimitSize || r.limit < 1 {
 		return apiutil.ErrLimitSize
 	}
 
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 
@@ -13,6 +14,8 @@ import (
 )
 
 var errStatusFilterUnsupported = errors.New("status filter is not supported")
+
+const maxMetadataBytes = 65536
 
 type taskReq struct {
 	task.Task `json:",inline"`
@@ -39,6 +42,16 @@ func (t *taskReq) validate() error {
 
 	if t.Broadcast && t.PropletID != "" {
 		return fmt.Errorf("%w: broadcast and proplet_id are mutually exclusive", pkgerrors.ErrInvalidValue)
+	}
+
+	if len(t.Metadata) > 0 {
+		b, err := json.Marshal(t.Metadata)
+		if err != nil {
+			return fmt.Errorf("invalid metadata: %w", err)
+		}
+		if len(b) > maxMetadataBytes {
+			return errors.New("metadata exceeds 64KB limit")
+		}
 	}
 
 	return nil
@@ -139,6 +152,16 @@ func (e listEntityReq) validate() error {
 	default:
 		return pkgerrors.ErrInvalidValue
 	}
+}
+
+type listTasksReq struct {
+	offset         uint64
+	limit          uint64
+	metadataFilter map[string]string
+}
+
+func (r *listTasksReq) validate() error {
+	return nil
 }
 
 type metricsReq struct {
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -156,9 +156,9 @@ func (e listEntityReq) validate() error {
 }
 
 type listTasksReq struct {
-	offset         uint64
-	limit          uint64
-	metadataFilter map[string]string
+	offset   uint64
+	limit    uint64
+	metadata map[string]any
 }
 
 func (r *listTasksReq) validate() error {
--- a/manager/api/transport.go
+++ b/manager/api/transport.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/absmach/magistrala"
@@ -82,7 +83,7 @@ func MakeHandler(svc manager.Service, logger *slog.Logger, instanceID string) ht
 		), "create-task").ServeHTTP)
 		r.Get("/", otelhttp.NewHandler(kithttp.NewServer(
 			listTasksEndpoint(svc),
-			decodeListEntityReq(withoutStatusFilter),
+			decodeListTasksReq,
 			api.EncodeResponse,
 			opts...,
 		), "list-tasks").ServeHTTP)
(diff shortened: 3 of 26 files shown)


JeffMboya:
## Updated metadata size limit: 64KB → 1MB

Pushed in 63e4213.

```go
const maxMetadataBytes = 1048576 // 1MB
```

**Why any limit at all?**

The `metadata` field gets serialized into three places that all have practical size constraints:

1. **MQTT dispatch** — on `StartTask`, the manager publishes the full task struct (including metadata) to the proplet over MQTT. The proplet's broker config caps packets at `mqtt_max_packet_size: 10485760` (10MB). A metadata blob that is large enough starts eating into that budget alongside the wasm binary, which can be several MB on its own.

2. **Database rows** — metadata is stored as a `JSONB` column in Postgres and a `TEXT`/`BLOB` in SQLite and Badger. There is no hard column-level limit, but very large JSON blobs increase row size, slow down list queries (every task row is scanned for metadata filter lookups), and inflate index entries.

3. **API responses** — list-tasks returns the full metadata for every task in the page. A 10MB metadata field on 100 tasks in one response is a 1GB payload.

**Why 1MB specifically?**

It is still arbitrary — there is no measurement behind it. 64KB was too tight for real use cases (structured ML configs, nested feature vectors, serialized schemas). 1MB is a common upper bound for "rich structured configuration" without crossing into "someone is storing a model weights file in metadata". If a concrete use case hits 1MB, the limit can be raised again or made configurable via env var.

manager/api/endpoint.go
```diff
@@ -243,22 +244,35 @@ func stopJobEndpoint(svc manager.Service) endpoint.Endpoint {
 
 func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 	return func(ctx context.Context, request any) (any, error) {
-		req, ok := request.(listEntityReq)
-		if !ok {
+		switch req := request.(type) {
+		case listEntityReq:
```
rodneyosodo: Follow supermq style, no need for switch statement here

manager/api/endpoint.go
```diff
@@ -243,22 +244,35 @@ func stopJobEndpoint(svc manager.Service) endpoint.Endpoint {
...
+			if err := req.validate(); err != nil {
+				return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
+			}
+			tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
```
rodneyosodo: Let's have one method, ListTasks, which takes in pageMetadata -> offset, limit, metadata

manager/api/requests.go
```diff
@@ -9,6 +11,8 @@ import (
 	apiutil "github.com/absmach/supermq/api/http/util"
 )
 
+const maxMetadataBytes = 1048576
```
rodneyosodo:
```suggestion
const maxMetadataBytes = 1048576  // 1MB
```

manager/api/requests.go
```diff
@@ -105,6 +119,16 @@ func (e *listEntityReq) validate() error {
 	return nil
 }
 
+type listTasksReq struct {
+	offset         uint64
+	limit          uint64
+	metadataFilter map[string]string
+}
+
+func (r *listTasksReq) validate() error {
```
rodneyosodo: We should prevent filtering with a huge limit

manager/api/transport.go
```diff
@@ -318,6 +319,54 @@ func decodeListEntityReq(_ context.Context, r *http.Request) (any, error) {
...
+		return nil, errors.Join(apiutil.ErrValidation, err)
+	}
+
+	l, err := apiutil.ReadNumQuery[uint64](r, api.LimitKey, api.DefLimit)
+	if err != nil {
+		return nil, errors.Join(apiutil.ErrValidation, err)
+	}
+
+	filter := map[string]string{}
+	for key, vals := range r.URL.Query() {
```
rodneyosodo: Kindly check how supermq does it

pkg/storage/badger/tasks.go
```diff
@@ -17,13 +18,24 @@ func NewTaskRepository(db *Database) TaskRepository {
 	return &taskRepo{db: db}
 }
 
+func metaIdxKey(k, v, id string) []byte {
```
rodneyosodo, line 23: Check how supermq does it. For update it is usually a full update

rodneyosodo (changes requested):

manager/api/endpoint.go
JeffMboya: Replaced switch with direct type assertion following supermq style.

manager/api/endpoint.go
JeffMboya: Consolidated into single `ListTasks(ctx, PageMetadata)` accepting offset, limit, metadata.

manager/api/requests.go
JeffMboya: Suggestion applied — `// 1MB` comment added.

manager/api/requests.go
JeffMboya: Added `limit > api.MaxLimitSize || limit < 1` guard in `listTasksReq.validate()`.

manager/api/transport.go
JeffMboya: Updated to use `apiutil.ReadMetadataQuery` and `apiutil.ReadNumQuery` per supermq style.

pkg/storage/badger/tasks.go
JeffMboya: Update is now a full overwrite — de-indexes old entry, writes new, re-indexes.

pkg/storage/repository.go
```diff
@@ -14,6 +14,7 @@ type TaskRepository interface {
 	Get(ctx context.Context, id string) (task.Task, error)
 	Update(ctx context.Context, t task.Task) error
 	List(ctx context.Context, offset, limit uint64) ([]task.Task, uint64, error)
+	ListByMetadataFilter(ctx context.Context, filter task.Metadata, offset, limit uint64) ([]task.Task, uint64, error)
```
rodneyosodo: Why not just have list which takes in offset, limit or metadata

rodneyosodo (changes requested):

pkg/storage/repository.go
JeffMboya: Done — merged into a single `List(ctx, filter, offset, limit)` that branches internally on whether the filter is empty.

manager/service.go
```diff
@@ -986,6 +948,24 @@ func (svc *service) RecoverInterruptedTasks(ctx context.Context) error {
...
 
+func (svc *service) updateCronTask(ctx context.Context, t task.Task) {
+	if svc.cronScheduler == nil {
+		return
+	}
+
+	if t.Schedule != "" {
+		if err := svc.cronScheduler.ScheduleTask(ctx, t.ID); err != nil {
+			svc.logger.WarnContext(ctx, "failed to reschedule task in cron scheduler", "error", err, "task_id", t.ID)
+		}
```
rodneyosodo, line 1053: **Bug: `Broadcast` field dropped from payload**

The original `map[string]any` payload included `"broadcast": t.Broadcast`, but `startPayload` has no `Broadcast` field. Proplets will never receive the broadcast flag — this silently breaks broadcast task dispatch.

```go
// Add to startPayload struct:
Broadcast bool `json:"broadcast"`

// And populate it:
Broadcast: t.Broadcast,
```

rodneyosodo (commented):
Overall the metadata feature is well-structured across storage backends, but the PR bundles several unrelated refactors that introduce critical regressions. See inline comments for details.

manager/service.go
```diff
@@ -1715,24 +1695,38 @@ func (svc *service) persistTaskBeforeStart(ctx context.Context, t *task.Task) er
 	return svc.taskRepo.Update(ctx, *t)
 }
 
+type startPayload struct {
+	ID                string                     `json:"id"`
```
rodneyosodo, line 1798: **Bug: `Broadcast` field is missing from `startPayload`**

The original `map[string]any` payload explicitly included `"broadcast": t.Broadcast`, but the new struct has no `Broadcast` field. This silently drops the broadcast flag — proplets will never receive it, breaking broadcast task dispatch.

```go
type startPayload struct {
    // ... existing fields ...
    Broadcast bool `json:"broadcast"`
}
```

manager/service.go
```diff
@@ -1715,24 +1695,38 @@ func (svc *service) persistTaskBeforeStart(ctx context.Context, t *task.Task) er
...
+	ImageURL          string                     `json:"image_url,omitempty"`
+	File              []byte                     `json:"file,omitempty"`
+	Inputs            task.FlexStrings           `json:"inputs,omitempty"`
+	CLIArgs           []string                   `json:"cli_args"`
+	Daemon            bool                       `json:"daemon"`
+	Env               map[string]string          `json:"env,omitempty"`
+	Encrypted         bool                       `json:"encrypted"`
+	KBSResourcePath   string                     `json:"kbs_resource_path,omitempty"`
+	MonitoringProfile *proplet.MonitoringProfile `json:"monitoringProfile,omitempty"`
+	PropletID         string                     `json:"proplet_id,omitempty"`
```
rodneyosodo, line 1811: **Bug: JSON tag renamed from `monitoring_profile` to `monitoringProfile`**

The original map used `"monitoring_profile"` as the key. The new struct tag is `json:"monitoringProfile,omitempty"`. Any proplet or consumer deserializing this field by name will silently get a zero value — a wire-breaking change with no migration path.

Keep the original tag: `json:"monitoring_profile,omitempty"`

manager/service.go
```diff
@@ -519,85 +520,46 @@ func (svc *service) GetTask(ctx context.Context, taskID string) (task.Task, erro
...
 
-	if err := svc.taskRepo.Update(ctx, dbT); err != nil {
+	if err := svc.taskRepo.Update(ctx, t); err != nil {
 		return task.Task{}, err
 	}
 
-	if !scheduleChanged || svc.cronScheduler == nil {
-		return dbT, nil
-	}
+	svc.updateCronTask(ctx, t)
```
rodneyosodo, line 655: **Unnecessary cron churn: `updateCronTask` always called**

Previously `scheduleChanged` guarded this block, so cron operations only ran when the schedule actually changed. Now `updateCronTask` is called on every update, which means `UnscheduleTask` fires for tasks that never had a schedule. This adds unnecessary overhead and log noise.

Consider restoring the guard:
```go
if scheduleChanged {
    svc.updateCronTask(ctx, t)
}
```

manager/api/requests.go
```diff
@@ -1,14 +1,19 @@
...
 
 	apiutil "github.com/absmach/magistrala/api/http/util"
+	"github.com/absmach/propeller/pkg/api"
 	"github.com/absmach/propeller/pkg/cron"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
 	"github.com/absmach/propeller/pkg/task"
 )
 
+const maxMetadataBytes = 1048576 // 1MB
+
```
rodneyosodo, line 20: **Size limit inconsistency: constant is 1 MB but PR description says 64 KB**

`maxMetadataBytes = 1048576` is 1 MB. The PR description states "must be valid JSON and under 64KB". Pick one limit and make the constant, the comment, and the error message consistent.

manager/manager.go
```diff
@@ -21,7 +22,7 @@ type Service interface {
 	ListJobs(ctx context.Context, offset, limit uint64) (JobPage, error)
 	StartJob(ctx context.Context, jobID string) error
 	StopJob(ctx context.Context, jobID string) error
-	ListTasks(ctx context.Context, offset, limit uint64) (task.TaskPage, error)
+	ListTasks(ctx context.Context, pm sdk.PageMetadata) (task.TaskPage, error)
 	UpdateTask(ctx context.Context, task task.Task) (task.Task, error)
```
rodneyosodo, line 33: **Architecture: core `Service` interface should not import `pkg/sdk`**

`pkg/sdk` is a client-facing wrapper. Having the core domain interface (`manager.Service`) depend on it creates an awkward inward dependency and forces all middleware, mocks, and the service itself to import the SDK package.

`PageMetadata` (or a domain-local equivalent with just `Offset`, `Limit`, `Metadata`) should live in `pkg/api` or be defined inline in the `manager` package.

pkg/storage/sqlite/tasks.go
```diff
@@ -181,23 +198,48 @@ func (r *taskRepo) Update(ctx context.Context, t task.Task) error {
 	return nil
 }
 
-func (r *taskRepo) List(ctx context.Context, offset, limit uint64) ([]task.Task, uint64, error) {
-	var total uint64
-	err := r.db.GetContext(ctx, &total, "SELECT COUNT(*) FROM tasks")
+func (r *taskRepo) List(ctx context.Context, filter task.Metadata, offset, limit uint64) ([]task.Task, uint64, error) {
+	whereClause, args, err := buildSQLiteMetadataWhere(filter)
 	if err != nil {
```
rodneyosodo, line 203: **SQL injection risk: metadata key interpolated into query string**

`buildSQLiteMetadataWhere` validates keys against `sqliteMetadataKeyRe` then writes them directly into the SQL via `sb.WriteString(k)`. Values are correctly parameterized (`?`), but keys are not.

The regex allowlist (`^[a-zA-Z0-9._\-]+$`) is tight enough to prevent injection today, but this pattern is fragile — a future maintainer might loosen the regex without realising the SQL impact. Prefer a parameterized approach, or at minimum add a comment explaining why direct interpolation is safe here.

pkg/storage/memory_adapter.go
```diff
@@ -104,9 +117,135 @@ func (r *memoryTaskRepo) ListByJobID(ctx context.Context, jobID string) ([]task.
...
+			delete(r.metaIdx[k], s)
+		}
+		if len(r.metaIdx[k]) == 0 {
+			delete(r.metaIdx, k)
+		}
+	}
+}
+
+func (r *memoryTaskRepo) listAll(ctx context.Context, offset, limit uint64) ([]task.Task, uint64, error) {
+	data, total, err := r.storage.List(ctx, offset, limit)
```
rodneyosodo, line 176: **Potential race: `listAll` does not hold `r.mu`**

`Create`, `Update`, and `Delete` all acquire `r.mu` before mutating state, but `listAll` calls `r.storage.List` without holding any lock. If the underlying `Storage` implementation is not independently thread-safe, concurrent writes and reads will race.

Add `r.mu.RLock() / r.mu.RUnlock()` at the top of `listAll`, or confirm that `Storage.List` is safe to call concurrently with mutations.

manager/service.go
```diff
@@ -519,85 +520,46 @@ func (svc *service) GetTask(ctx context.Context, taskID string) (task.Task, erro
...
-	if t.Inputs != nil {
-		dbT.Inputs = t.Inputs
-	}
-	if t.File != nil {
-		dbT.File = t.File
-	}
-	if t.Priority != 0 {
-		dbT.Priority = t.Priority
-	}
+	t.UpdatedAt = time.Now()
```
rodneyosodo, line 607: **Breaking change: partial updates are no longer supported**

Previously `UpdateTask` fetched the DB record and merged only non-zero fields (name, inputs, file, priority, schedule) — a PATCH-style semantic. Now it passes the caller's `t` directly to the repo, so any field left at its zero value will overwrite the stored data.

For example, a client that only intends to update `t.Schedule` must now populate every other field too, or risk zeroing `Inputs`, `File`, `Priority`, etc.

If this is intentional (PUT semantics), document it and update the HTTP handler accordingly. If not, restore the merge logic.

pkg/storage/badger/tasks.go
```diff
@@ -50,14 +62,113 @@ func (r *taskRepo) Update(ctx context.Context, t task.Task) error {
...
+	})
+}
+
+func (r *taskRepo) Delete(ctx context.Context, id string) error {
+	key := []byte("task:" + id)
+	err := r.db.updateTxn(func(txn *badgerdb.Txn) error {
+		item, err := txn.Get(key)
+		if err != nil {
+			return txn.Delete(key)
+		}
```
rodneyosodo, line 125: **Error handling: deleting a key after `txn.Get` fails is incorrect**

When `txn.Get` returns `ErrKeyNotFound`, the code falls through and calls `txn.Delete(key)` — deleting something that does not exist. This should either return `nil` (idempotent delete) or return the original `ErrKeyNotFound` error. Deleting unconditionally on any `Get` error (including transient errors) is also risky.

```go
item, err := txn.Get(key)
if errors.Is(err, badgerdb.ErrKeyNotFound) {
    return nil // already gone, nothing to deindex
}
if err != nil {
    return err
}
```

pkg/storage/sqlite/init.go
```diff
@@ -253,6 +253,15 @@ func (db *Database) Migrate() error {
 					`ALTER TABLE tasks DROP COLUMN broadcast`,
 				},
 			},
+			{
+				Id: "5_add_task_metadata",
+				Up: []string{
```
rodneyosodo, line 259: **Performance: SQLite metadata column has no index**

Postgres adds a GIN index (`idx_tasks_metadata`) for efficient JSONB containment queries. SQLite only adds the column — `json_extract` in `buildSQLiteMetadataWhere` will do a full table scan on every filtered list request.

SQLite supports expression/generated-column indexes that could help, or at minimum document the performance limitation.

pkg/storage/badger/tasks.go
```diff
@@ -50,14 +62,113 @@ func (r *taskRepo) Update(ctx context.Context, t task.Task) error {
...
+	}
+
+	return nil
+}
+
+func (r *taskRepo) indexTaskTxn(txn *badgerdb.Txn, t task.Task) error {
+	for k, v := range t.Metadata {
+		s, ok := v.(string)
+		if !ok {
+			continue
```
rodneyosodo: **Silent data loss: non-string metadata values are not indexed**

When a metadata value is not a `string` (e.g. `int`, `bool`, `float64`), it is silently skipped during indexing and de-indexing — and by extension, during filter evaluation in `filterIDsByMetadata`. A caller filtering by `{"count": 5}` will always get an empty result with no error.

The same pattern exists in the memory repo. Either:
- Reject non-string values at the API/validation layer (document that metadata values must be strings), or
- Support non-string values by converting them to a canonical string representation before indexing.

manager/service.go
JeffMboya: Fixed: changed JSON tag back to monitoring_profile,omitempty.

manager/service.go
JeffMboya: Fixed: restored scheduleChanged guard; updateCronTask only called when schedule changes.

manager/api/requests.go
JeffMboya: Constant, comment, and error message are all 1MB — already consistent.

manager/manager.go
JeffMboya: Fixed: defined PageMetadata in manager package; removed pkg/sdk from Service interface.

pkg/storage/sqlite/tasks.go
JeffMboya: Added comment explaining why key interpolation is safe via regex allowlist.

pkg/storage/memory_adapter.go
JeffMboya: Fixed: added r.mu.RLock()/RUnlock() at the top of listAll.

manager/service.go
JeffMboya: Fixed: restored PATCH semantics — fetch, merge non-zero fields, then save.

pkg/storage/badger/tasks.go
JeffMboya: Fixed: return nil on ErrKeyNotFound; return err on other txn.Get failures.

pkg/storage/sqlite/init.go
JeffMboya: Added partial index on metadata IS NOT NULL; added comment about full-scan limitation.

pkg/storage/badger/tasks.go
JeffMboya: Fixed: use fmt.Sprint for non-string values in badger and memory index/filter.

manager/service.go
JeffMboya: Already fixed in f546d0b — Broadcast field added to startPayload struct.

manager/service.go
```diff
@@ -1715,24 +1724,40 @@ func (svc *service) persistTaskBeforeStart(ctx context.Context, t *task.Task) er
 	return svc.taskRepo.Update(ctx, *t)
 }
 
+type startPayload struct {
```
rodneyosodo, line 1797: **`startPayload` omits `Metadata` — tasks dispatched to proplets will have metadata stripped.**

Every field from `task.Task` was ported to `startPayload` except `Metadata`. If metadata is used to configure task execution at the proplet side (e.g. resource limits, routing hints), it will be silently dropped on dispatch.

If metadata is intentionally manager-only, add a comment here to explain that decision:
```go
// Metadata is intentionally excluded: it is used for filtering in the manager
// and is not needed by the proplet runtime.
```
If it should be sent, add the field:
```go
Metadata task.Metadata `json:"metadata,omitempty"`
```

pkg/storage/memory_adapter.go
```diff
@@ -104,9 +117,137 @@ func (r *memoryTaskRepo) ListByJobID(ctx context.Context, jobID string) ([]task.
...
+		}
+		if len(matchIDs) == 0 {
+			return nil, false
+		}
+	}
+
+	return matchIDs, true
+}
+
+func (r *memoryTaskRepo) listByMetadata(ctx context.Context, filter task.Metadata, offset, limit uint64) ([]task.Task, uint64, error) {
```
rodneyosodo, line 227: **Non-deterministic pagination: map iteration order is random, so `offset`/`limit` returns different tasks across calls.**

After collecting matching IDs from `filterIDsByMetadata` (a `map[string]struct{}`), the IDs are iterated in random order, tasks are appended to a slice, and then paginated. A caller requesting page 2 may receive results that overlap with page 1.

**Fix:** Sort the result slice before slicing:
```go
tasks := make([]task.Task, 0, len(matchIDs))
for id := range matchIDs {
    // ... fetch task ...
    tasks = append(tasks, t)
}
slices.SortFunc(tasks, func(a, b task.Task) int {
    return strings.Compare(a.ID, b.ID)
})
total := uint64(len(tasks))
if offset >= total {
    return []task.Task{}, total, nil
}
return tasks[offset:min(offset+limit, total)], total, nil
```
The same issue applies to `pkg/storage/badger/tasks.go:198`.

pkg/storage/sqlite/tasks.go
```diff
@@ -181,23 +198,54 @@ func (r *taskRepo) Update(ctx context.Context, t task.Task) error {
...
 	return tasks, total, nil
 }
 
+// buildSQLiteMetadataWhere builds a WHERE clause for metadata filtering.
+// Keys are interpolated directly into the SQL because sqliteMetadataKeyRe
+// restricts them to [a-zA-Z0-9._\-]+, which contains no SQL metacharacters.
+// Values are always bound as parameters. Metadata filtering performs a full
+// table scan since SQLite does not support expression indexes on json_extract
+// for arbitrary key sets.
+func buildSQLiteMetadataWhere(filter task.Metadata) (clause string, args []any, err error) {
```
rodneyosodo, line 229: **SQLite type mismatch: non-string metadata values silently return 0 rows.**

`json_extract(metadata, '$.key')` returns a typed SQLite value (integer, real, text, etc.) from the stored JSON. The bound parameter `v` is a Go `any` — when it comes from an HTTP query string it will be a `string`. SQLite's `=` comparison between an integer column value and a string parameter fails silently (returns false), so filtering by `?metadata={"count":5}` returns nothing even when tasks have `{"count": 5}` stored.

The Postgres backend avoids this because `@>` is a JSONB containment operator that compares typed values after JSON parsing.

**Fix option 1:** Normalize all stored values to strings at write time (consistent with Badger's `metaValueString`), and document that metadata values are always compared as strings.

**Fix option 2:** JSON-encode the filter value before binding so SQLite can use `json()` for type-aware comparison:
```go
vJSON, err := json.Marshal(v)
if err != nil {
    return "", nil, fmt.Errorf("invalid metadata value: %w", err)
}
// expression: json_extract(metadata, '$."key"') = json(?)
// bound param: string(vJSON)  → 5 not "5", true not "true"
args = append(args, string(vJSON))
// expression becomes:
sb.WriteString(`') = json(?)`)
```

pkg/storage/sqlite/init.go
```diff
@@ -253,6 +253,17 @@ func (db *Database) Migrate() error {
...
+					`ALTER TABLE tasks ADD COLUMN metadata TEXT`,
+					`CREATE INDEX IF NOT EXISTS idx_tasks_metadata ON tasks(metadata) WHERE metadata IS NOT NULL`,
```
rodneyosodo: **This B-tree index on the raw `metadata TEXT` column won't be used by `json_extract(metadata, '$.key')` queries** — SQLite's query planner can't match a B-tree index on the column to an expression index on a JSON path.

The comment in `buildSQLiteMetadataWhere` correctly documents the full-table-scan behaviour, but the index migration still runs on every deployment and wastes space without benefit.

**Options:**
1. Remove the index entirely and rely on the documented full-table-scan (honest and cheap).
2. Replace with a partial index that at least makes unfiltered `WHERE metadata IS NOT NULL` counts faster:
   ```sql
   CREATE INDEX IF NOT EXISTS idx_tasks_has_metadata ON tasks(id) WHERE metadata IS NOT NULL
   ```
3. For production use, consider using a separate metadata search table as a proper index structure.

manager/api/requests.go
```diff
@@ -1,14 +1,19 @@
...
 	"fmt"
 
 	apiutil "github.com/absmach/magistrala/api/http/util"
+	"github.com/absmach/propeller/pkg/api"
 	"github.com/absmach/propeller/pkg/cron"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
 	"github.com/absmach/propeller/pkg/task"
 )
 
+const maxMetadataBytes = 1048576 // 1MB
```
rodneyosodo, line 19: **PR description says 64KB limit; code is 1MB.** One of them is wrong — align them before merging.

Also: the current validation marshals `t.Metadata` here to check its size, and then the storage layer marshals it again. If the intent is just a size guard, you can skip the marshal and check a rough upper bound on the number of map entries instead, or measure the size at the HTTP boundary before decoding.

manager/api/requests.go
```diff
@@ -105,6 +120,20 @@ func (e *listEntityReq) validate() error {
 	return nil
 }
 
+type listTasksReq struct {
+	offset   uint64
+	limit    uint64
+	metadata task.Metadata
+}
+
+func (r *listTasksReq) validate() error {
```
rodneyosodo: `listTasksReq.validate()` uses a pointer receiver `(r *listTasksReq)` while every other `validate()` in this file uses a value receiver. Inconsistency — change to `(r listTasksReq)` to match the pattern.

rodneyosodo (commented):
Overall solid work — the Badger indexing approach and memory adapter locking are well designed. Several correctness issues worth fixing before merge.

rodneyosodo (approved):
LGTM

PR #171, opened 2026-03-26
PROP-57 - Enable filtering of events and jobs

smithjilks:
# What type of PR is this?

PR #180, opened 2026-04-07
PROP-57 - Add status filter for proplets and jobs

### LINKED ISSUE #57: Feature: Enable Filtering of Active and Inactive Entities

Currently, we aren't able to filter proplets based on their status. For example active or inactive proplets

Implement filtering based on status for jobs and proplets so that I can list active proplets, inactive proplets, running jobs, finished jobs etc

Nice-to-have

### Anything else?

_No response_

JeffMboya:
## What type of PR is this?

Feature - status-based filtering for proplets and jobs.

## What does this do?

- `GET /proplets?status=active|inactive` filters by liveness
- `GET /jobs?status=pending|running|completed|failed` filters by state
- `GET /tasks` returns 400 if `?status=` is passed
- SDK: `ListProplets` and `ListJobs` accept optional status param
- Invalid status values return 400, not 500

- Resolves #57. Supersedes #171 from @smithjilks.

Yes - service-level and filter-pagination tests added.

No - API is self-describing via query param validation errors.

### DIFF
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -20,7 +20,7 @@ func listPropletsEndpoint(svc manager.Service) endpoint.Endpoint {
 			return listpropletResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
 
-		proplets, err := svc.ListProplets(ctx, req.offset, req.limit)
+		proplets, err := svc.ListProplets(ctx, req.offset, req.limit, req.status)
 		if err != nil {
 			return listpropletResponse{}, err
 		}
@@ -188,7 +188,7 @@ func listJobsEndpoint(svc manager.Service) endpoint.Endpoint {
 			return listJobResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
 
-		jobs, err := svc.ListJobs(ctx, req.offset, req.limit)
+		jobs, err := svc.ListJobs(ctx, req.offset, req.limit, req.status)
 		if err != nil {
 			return listJobResponse{}, err
 		}
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -250,6 +250,9 @@ func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
+		if req.status != "" {
+			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, pkgerrors.ErrInvalidData)
+		}
 
 		tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
 		if err != nil {
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -251,7 +251,7 @@ func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
 		if req.status != "" {
-			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, pkgerrors.ErrInvalidData)
+			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, pkgerrors.ErrInvalidValue)
 		}
 
 		tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
--- a/manager/api/endpoint.go
+++ b/manager/api/endpoint.go
@@ -250,9 +250,6 @@ func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
-		if req.status != "" {
-			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, pkgerrors.ErrInvalidValue)
-		}
 
 		tasks, err := svc.ListTasks(ctx, req.offset, req.limit)
 		if err != nil {
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -1,13 +1,17 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 
 	apiutil "github.com/absmach/magistrala/api/http/util"
 	"github.com/absmach/propeller/pkg/cron"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
+	"github.com/absmach/propeller/pkg/proplet"
 	"github.com/absmach/propeller/pkg/task"
 )
+
+var errStatusFilterUnsupported = errors.New("status filter is not supported")
 
 type taskReq struct {
 	task.Task `json:",inline"`
@@ -97,13 +101,37 @@ func (e *entityReq) validate() error {
 	return nil
 }
 
+type listEntityStatus uint8
+
+const (
+	withoutStatusFilter listEntityStatus = iota
+	propletStatusFilter
+	jobStatusFilter
+)
+
 type listEntityReq struct {
 	offset, limit uint64
 	status        string
+	statusFilter  listEntityStatus
 }
 
-func (e *listEntityReq) validate() error {
-	return nil
+func (e listEntityReq) validate() error {
+	if e.status == "" {
+		return nil
+	}
+
+	switch e.statusFilter {
+	case withoutStatusFilter:
+		return errStatusFilterUnsupported
+	case propletStatusFilter:
+		_, err := proplet.ToStatus(e.status)
+		return err
+	case jobStatusFilter:
+		_, err := task.ToJobStatus(e.status)
+		return err
+	default:
+		return pkgerrors.ErrInvalidValue
+	}
 }
 
 type metricsReq struct {
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -125,9 +125,11 @@ func (e listEntityReq) validate() error {
 		return errStatusFilterUnsupported
 	case propletStatusFilter:
 		_, err := proplet.ToStatus(e.status)
+
 		return err
 	case jobStatusFilter:
 		_, err := task.ToJobStatus(e.status)
+
 		return err
 	default:
 		return pkgerrors.ErrInvalidValue
--- a/manager/api/requests.go
+++ b/manager/api/requests.go
@@ -99,6 +99,7 @@ func (e *entityReq) validate() error {
 
 type listEntityReq struct {
 	offset, limit uint64
+	status        string
 }
 
 func (e *listEntityReq) validate() error {
--- a/manager/service.go
+++ b/manager/service.go
@@ -432,10 +432,6 @@ func (svc *service) ListJobs(ctx context.Context, offset, limit uint64, status s
 	})
 
 	if status != "" {
-		// ComputeJobState only ever returns Pending, Running, Completed, or Failed.
-		// Skipped and Interrupted task states are collapsed: Interrupted → Failed,
-		// Scheduled → Running, all-Skipped → Completed. No job summary can carry
-		// any other state, so the map below is exhaustive.
 		statusStateMap := map[string]task.State{
 			JobStatusPending:   task.Pending,
 			JobStatusRunning:   task.Running,
--- a/manager/service.go
+++ b/manager/service.go
@@ -126,7 +126,7 @@ func (svc *service) GetProplet(ctx context.Context, propletID string) (proplet.P
 
 func (svc *service) ListProplets(ctx context.Context, offset, limit uint64, status string) (proplet.PropletPage, error) {
 	if status != "" && status != PropletStatusActive && status != PropletStatusInactive {
-		return proplet.PropletPage{}, fmt.Errorf("invalid proplet status filter %q: must be %q, %q, or empty", status, PropletStatusActive, PropletStatusInactive)
+		return proplet.PropletPage{}, fmt.Errorf("%w: proplet status must be %q, %q, or empty, got %q", pkgerrors.ErrInvalidValue, PropletStatusActive, PropletStatusInactive, status)
 	}
 
 	if status == "" {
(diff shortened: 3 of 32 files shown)


manager/service.go
```diff
@@ -39,6 +39,14 @@ const (
 	ExecutionModeConfigurable = "configurable"
 	EnvJobExecutionMode       = "JOB_EXECUTION_MODE"
 	shutdownTaskStopWait      = 200 * time.Millisecond
+
+	PropletStatusActive   = "active"
```
rodneyosodo: These status constants live in `service.go` alongside the service implementation. They would be more discoverable—and reusable by callers without importing the full manager package—if they lived in `pkg/proplet` (for `PropletStatus*`) and alongside `task.State` in `pkg/task` (for `JobStatus*`). At minimum, a dedicated `manager/const.go` would keep the implementation file focused.

manager/api/requests.go
```diff
@@ -99,6 +99,7 @@ func (e *entityReq) validate() error {
 
 type listEntityReq struct {
 	offset, limit uint64
+	status        string
```
rodneyosodo, line 114: The `status` field is added to the shared `listEntityReq` but `validate()` never inspects it. `offset` and `limit` are caught at the transport layer via `apiutil` helpers; it would be more consistent to validate `status` here with an allow-list check too. As-is, an invalid proplet/job status passes silently through transport and endpoint and is only caught deep in the service, while the tasks endpoint has its own ad-hoc guard in the endpoint function.

manager/api/endpoint.go
```diff
@@ -250,6 +250,9 @@ func listTasksEndpoint(svc manager.Service) endpoint.Endpoint {
 		if err := req.validate(); err != nil {
 			return listTaskResponse{}, errors.Join(apiutil.ErrValidation, err)
 		}
+		if req.status != "" {
```
rodneyosodo: This guard is correct, but its placement is inconsistent with how proplet and job status validation works (which happens in `service.go`). Consider centralising all status validation in `listEntityReq.validate()` with a per-entity allow-list, so all three list endpoints follow the same path from request parsing → validation → service call.

manager/service.go
```diff
@@ -397,6 +432,23 @@ func (svc *service) ListJobs(ctx context.Context, offset, limit uint64) (JobPage
 		}
 	})
 
+	if status != "" {
```
rodneyosodo, line 436: Job filtering happens entirely in memory after all tasks are fetched, deduplicated, and sorted. There is no DB-side push-down the way proplets have `ListByAlive`. Pagination and `total` are applied to the filtered slice (which is correct), but for large task stores this is O(N). This is structurally constrained by the current in-memory job aggregation model, but a comment here would help future readers understand why there is no storage-level `ListByState` call analogous to `ListByAlive`.

pkg/storage/sqlite/proplets.go
```diff
@@ -121,6 +121,46 @@ func (r *propletRepo) List(ctx context.Context, offset, limit uint64) ([]proplet
...
 }
 
+func (r *propletRepo) ListByAlive(ctx context.Context, offset, limit uint64, alive bool, since time.Time) ([]proplet.Proplet, uint64, error) {
+	tx, err := r.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", ErrDBQuery, err)
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	rows, err := tx.QueryContext(ctx, `SELECT id, name, task_count, alive, alive_history, metadata FROM proplets`)
```
rodneyosodo: This fetches **every row** from `proplets` with no `WHERE` clause and filters liveness in Go. The Postgres implementation correctly pushes the filter to SQL; SQLite supports JSON path expressions that can do the same:

```sql
SELECT ... FROM proplets
WHERE json_extract(alive_history, '$[#-1]') >= ?
```

Or at minimum add a `LIMIT` cap similar to `maxBadgerScan` in the Badger implementation. As written, a table with many proplets loads all rows into memory on every filtered list call.

rodneyosodo (commented):
Rebase with main

pkg/storage/badger/proplets.go
```diff
@@ -79,6 +79,36 @@ func (r *propletRepo) List(ctx context.Context, offset, limit uint64) ([]proplet
 	return proplets, total, nil
 }
 
+const maxBadgerScan uint64 = 100000
+
+func (r *propletRepo) ListByAlive(ctx context.Context, offset, limit uint64, alive bool, since time.Time) ([]proplet.Proplet, uint64, error) {
+	prefix := []byte("proplet:")
+	values, err := r.db.listWithPrefix(prefix, 0, maxBadgerScan)
```
rodneyosodo: The regular `List` path passes the caller's `offset`/`limit` to `listWithPrefix` for key-range scanning, but `ListByAlive` overrides those with `0`/`maxBadgerScan` and paginates in Go. This is intentional since liveness cannot be pushed into Badger key iteration, but a deployment with >100 000 proplets will silently return truncated totals. Consider documenting this cap or returning an error when `filteredTotal` equals `maxBadgerScan` so callers know the result may be incomplete.

manager/service_proplet_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+			page, err := svc.ListProplets(ctx, tc.offset, tc.limit, manager.PropletStatusActive)
+			require.NoError(t, err)
+			assert.Equal(t, tc.expectedTotal, page.Total)
+			assert.Len(t, page.Proplets, tc.expectedLen)
+		})
+	}
+}
+
+func boolPtr(b bool) *bool { return &b }
```
rodneyosodo: `boolPtr` is used only in this file (two call sites). Either inline it at the call sites or move it to a shared test-helper location if other test files will need it — one-off private helpers like this tend to get quietly duplicated across the package.

manager/service.go
JeffMboya: Moved — `PropletStatus*` now lives in `pkg/proplet` as `ActiveStatus`/`InactiveStatus`; `JobStatus*` moved to `pkg/task` alongside `task.State`. No constants remain in `service.go`.

manager/api/requests.go
JeffMboya: Fixed — `validate()` now inspects `statusFilter` and calls `proplet.ToStatus` or `task.ToJobStatus` with a full allow-list check before hitting the service.

manager/api/endpoint.go
JeffMboya: All status validation is now centralised in `listEntityReq.validate()`. The tasks endpoint returns `errStatusFilterUnsupported` through the same path; no ad-hoc endpoint guard remains.

manager/service.go
JeffMboya: Comment added explaining job state is derived from aggregated tasks in memory rather than stored independently, making a storage-level push-down structurally impossible.

pkg/storage/sqlite/proplets.go
JeffMboya: `ListByAlive` now uses a SQL `WHERE` clause with `json_extract(alive_history, '$[#-1]')` and a `COUNT(*)` for total — no full table scan.

pkg/storage/badger/proplets.go
JeffMboya: Now returns an explicit error when the unfiltered keyspace exceeds `maxBadgerScan`, so callers are never silently given incomplete results.

manager/service_proplet_test.go
JeffMboya: Removed — `boolPtr` no longer exists in the test files.

JeffMboya:
Done — branch has been rebased on top of `upstream/main`.

pkg/storage/sqlite/proplets.go
```diff
@@ -121,6 +121,51 @@ func (r *propletRepo) List(ctx context.Context, offset, limit uint64) ([]proplet
...
+func (r *propletRepo) ListByAlive(ctx context.Context, offset, limit uint64, alive bool, since time.Time) ([]proplet.Proplet, uint64, error) {
+	tx, err := r.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", ErrDBQuery, err)
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	sinceArg := since.Format(time.RFC3339Nano)
+
+	whereClause := `WHERE alive_history IS NOT NULL AND json_array_length(alive_history) > 0 AND unixepoch(json_extract(alive_history, '$[#-1]')) >= unixepoch(?)`
```
rodneyosodo: **Bug: `$[#-1]` requires SQLite ≥ 3.38.0 (Feb 2022) — older builds silently return `NULL`, making every proplet appear inactive.**

The `$[#-1]` last-element shorthand was introduced in SQLite 3.38.0. On Ubuntu 20.04 (ships 3.31) or any older distro, `json_extract(alive_history, '$[#-1]')` returns `NULL`, so the `>=`/`<` comparisons are always false/true and every proplet is classified as inactive regardless of its actual heartbeat history. No error is raised.

**Fix:** Replace `$[#-1]` with an explicit index using `json_array_length`:

```sql
-- active
WHERE alive_history IS NOT NULL
  AND json_array_length(alive_history) > 0
  AND unixepoch(json_extract(alive_history,
        '$[' || (json_array_length(alive_history) - 1) || ']')) >= unixepoch(?)

-- inactive
WHERE alive_history IS NULL
   OR json_array_length(alive_history) = 0
   OR unixepoch(json_extract(alive_history,
        '$[' || (json_array_length(alive_history) - 1) || ']')) < unixepoch(?)
```

This works on all SQLite versions ≥ 3.7 (2010). Alternatively, document a minimum SQLite 3.38 requirement in the README/deployment notes.

pkg/proplet/proplet.go
```diff
@@ -1,8 +1,48 @@
...
+	ActiveStatus Status = iota
+	InactiveStatus
+)
+
+const (
+	AliveTimeout = 10 * time.Second
+
+	Active   = "active"
+	Inactive = "inactive"
+	Unknown  = "unknown"
```
rodneyosodo: **`Unknown` is exported but passing it as a `?status=` query value returns 400**, which is confusing for API consumers who discover the constant.

`ToStatus("unknown")` hits the `default` case and returns `ErrInvalidStatus`, so callers reading the exported constant and passing it verbatim get a validation error with no obvious hint.

**Two options:**
1. Unexport it: `unknown = "unknown"` — it's only used as the default `String()` return for `Status.String()`, which is displayed in logs/errors, not passed as API input.
2. Make it a valid filter alias that means "no filter" (equivalent to `""`), matching how `""` works in `ListProplets`.

Option 1 is simpler and avoids the semantic ambiguity.

pkg/storage/badger/proplets.go
```diff
@@ -79,6 +79,46 @@ func (r *propletRepo) List(ctx context.Context, offset, limit uint64) ([]proplet
...
 
+const maxBadgerScan uint64 = 100000
+
+func (r *propletRepo) ListByAlive(ctx context.Context, offset, limit uint64, alive bool, since time.Time) ([]proplet.Proplet, uint64, error) {
+	prefix := []byte("proplet:")
+	total, err := r.db.countWithPrefix(prefix)
+	if err != nil {
+		return nil, 0, err
+	}
+	if total > maxBadgerScan {
```
rodneyosodo, line 92: **Hard cap of 100K returns a 500 error at runtime, while memory/postgres/sqlite backends have no such limit.** An operator running Badger with >100K proplets hits this cap with no warning — and the behavior is inconsistent across backends, which breaks the abstraction.

**Suggested fix:** Replace the hard error with a warning log + best-effort scan, and document the cap at the `PropletRepository` interface level:

```go
const maxBadgerScan uint64 = 100_000

func (r *propletRepo) ListByAlive(...) ([]proplet.Proplet, uint64, error) {
    prefix := []byte("proplet:")
    total, err := r.db.countWithPrefix(prefix)
    if err != nil {
        return nil, 0, err
    }
    scanLimit := total
    if total > maxBadgerScan {
        // Log a warning but continue; results will be truncated.
        // TODO: consider an indexed approach for large deployments.
        scanLimit = maxBadgerScan
    }
    values, err := r.db.listWithPrefix(prefix, 0, scanLimit)
    // ...
}
```

Alternatively, document the 100K limit at the `PropletRepository` interface with a `// Note: Badger implementation caps filtered scans at 100K records` comment, so callers know what to expect.

rodneyosodo (commented):
Overall the design is clean and well-tested. A few issues worth fixing before merge — see inline comments.

pkg/storage/memory_adapter.go
```diff
@@ -153,6 +153,33 @@ func (r *memoryPropletRepo) List(ctx context.Context, offset, limit uint64) ([]p
 	return proplets, total, nil
 }
 
+func (r *memoryPropletRepo) ListByAlive(ctx context.Context, offset, limit uint64, alive bool, since time.Time) ([]proplet.Proplet, uint64, error) {
+	data, _, err := r.storage.List(ctx, 0, maxMemoryFetch)
```
rodneyosodo: `maxMemoryFetch = 100_000` is used here as the scan limit. If the in-memory store has more than 100K proplets, `ListByAlive` will silently return an incomplete filtered result — no error, just missing data.

This is the same cap as Badger, and the same concern: the memory backend should either be unbounded (since all data is already in memory — a full scan is cheap) or at minimum document the cap. Consider replacing `maxMemoryFetch` with a `math.MaxInt`-based fetch or just iterating the internal map directly, since there is no I/O cost.

manager/service.go
```diff
@@ -346,7 +370,18 @@ func (svc *service) GetJob(ctx context.Context, jobID string) ([]task.Task, erro
 	return svc.getJobTasks(ctx, jobID)
 }
 
-func (svc *service) ListJobs(ctx context.Context, offset, limit uint64) (JobPage, error) {
+func (svc *service) ListJobs(ctx context.Context, offset, limit uint64, status string) (JobPage, error) {
```
rodneyosodo, line 376: **`ListJobs` with a status filter fetches and aggregates *all* tasks before filtering — no early exit.** For a deployment with many jobs, `GET /jobs?status=pending` is as expensive as `GET /jobs` with no filter, because `listAllTasks` + sort + aggregate runs in full before the filtered slice is built.

This is an architectural constraint (job state is derived dynamically), and the comment explains why. But it should be documented as a known limitation in the issue/PR so consumers know the filter is presentational, not a performance optimisation. Consider adding a `// O(total tasks): job state is derived in-memory; status filter does not reduce storage I/O` comment here to make this explicit to future maintainers.

pkg/storage/sqlite/proplets.go
JeffMboya: Fixed. Replaced $[#-1] with '$[' || (json_array_length(alive_history) - 1) || ']' — works on SQLite ≥ 3.7.

pkg/proplet/proplet.go
JeffMboya: Unexported to unknown. It's only used as the default return value of Status.String() — not a valid API filter input.

pkg/storage/badger/proplets.go
JeffMboya: Replaced the hard error with a slog.Warn + best-effort truncation at the cap. Results are incomplete but the call no longer returns 500.

pkg/storage/memory_adapter.go
JeffMboya: Replaced maxMemoryFetch with math.MaxUint64 for all internal full-scan calls (offset=0, no overflow). Removed the constant.

manager/service.go
JeffMboya: Added an O(total tasks) doc comment explaining that the status filter is presentational and does not reduce storage I/O.

PR #182, opened 2026-04-08
PROP-88 -  Use Semantic Definition Format (SDF) for Data and Interactions of Things

### LINKED ISSUE #88: Feature: Use Semantic Definition Format (SDF) for Data and Interactions of Things

No

https://datatracker.ietf.org/doc/draft-ietf-asdf-sdf/

Must-have

### Anything else?

_No response_

JeffMboya:
# What type of PR is this?

This is a feature because it adds SDF (Semantic Definition Format) support for describing proplet data and interactions.

## What does this do?

Introduces a new `pkg/sdf` package that implements the [Semantic Definition Format (SDF)](https://datatracker.ietf.org/doc/draft-ietf-asdf-sdf/) for describing proplet data and interactions as standardized Thing descriptions.

- Adds `pkg/sdf` with `Document` and affordance types (`PropertyAfford`, `ActionAfford`, `EventAfford`, `ObjectAfford`, `ThingAfford`, `DataAfford`) following the SDF spec
- Adds `PropletDocument(p proplet.Proplet) sdf.Document` which builds a full SDF Thing description for a proplet, including properties (`id`, `name`, `alive`, `task_count`, `metadata`), actions (`start_task`, `stop_task`), events (`heartbeat`, `task_result`), and data schemas (`PropletMetadata`, `TaskDispatch`, `HeartbeatPayload`, `TaskResult`)
- Wires `GetPropletSDF` through the full manager stack: service interface, implementation, endpoint, HTTP transport (`GET /proplets/{id}/sdf`), logging/metrics/tracing middleware, and mock
- Exposes `GetPropletSDF(id string) (sdf.Document, error)` on the SDK interface

- Resolves #88

Yes. `pkg/sdf/proplet_test.go` contains table-driven tests covering:
- Correct `Info` fields (title contains proplet ID, non-empty version)
- All expected `sdfProperty`, `sdfAction`, `sdfEvent`, and `sdfData` keys on the `Proplet` thing
- `TaskResult.state` enum completeness (`Completed`, `Failed`, `Skipped`, `Interrupted`)
- `Minimum` pointer aliasing safety across `DataAfford` fields

The new `GET /proplets/{id}/sdf` endpoint and `GetPropletSDF` SDK method follow the existing patterns in the codebase. The SDK interface entry includes an inline usage example consistent with all other SDK methods.

### Notes

The SDF document is returned as a flat JSON object (fields of `sdf.Document` directly at the top level), consistent with how other response types embed their domain structs. The shape matches what the SDK `GetPropletSDF` unmarshals.

### DIFF
--- a/manager/api/transport_test.go
+++ b/manager/api/transport_test.go
@@ -1,0 +1,97 @@
+package api_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	managerapi "github.com/absmach/propeller/manager/api"
+	"github.com/absmach/propeller/manager/mocks"
+	pkgerrors "github.com/absmach/propeller/pkg/errors"
+	"github.com/absmach/propeller/pkg/sdf"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func newServer(t *testing.T) (*httptest.Server, *mocks.MockService) {
+	t.Helper()
+	svc := new(mocks.MockService)
+	handler := managerapi.MakeHandler(svc, slog.Default(), "test")
+
+	return httptest.NewServer(handler), svc
+}
+
+func TestGetPropletSDF(t *testing.T) {
+	t.Parallel()
+
+	validID := uuid.NewString()
+	validDoc := sdf.Document{
+		Info: sdf.Info{Title: "test", Version: "1.0"},
+	}
+
+	cases := []struct {
+		desc       string
+		propletID  string
+		svcDoc     sdf.Document
+		svcErr     error
+		wantStatus int
+		wantTitle  string
+	}{
+		{
+			desc:       "get SDF for existing proplet",
+			propletID:  validID,
+			svcDoc:     validDoc,
+			svcErr:     nil,
+			wantStatus: http.StatusOK,
+			wantTitle:  "test",
+		},
+		{
+			desc:       "get SDF for unknown proplet returns 404",
+			propletID:  uuid.NewString(),
+			svcDoc:     sdf.Document{},
+			svcErr:     pkgerrors.ErrNotFound,
+			wantStatus: http.StatusNotFound,
+		},
+		{
+			desc:       "get SDF with invalid proplet ID returns 500",
+			propletID:  "not-a-valid-uuid",
+			svcDoc:     sdf.Document{},
+			svcErr:     pkgerrors.ErrNotFound,
+			wantStatus: http.StatusNotFound,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+			ts, svc := newServer(t)
+			defer ts.Close()
+
+			if tc.propletID != "" {
+				svc.On("GetPropletSDF", mock.Anything, tc.propletID).Return(tc.svcDoc, tc.svcErr)
+			}
+
+			reqURL := fmt.Sprintf("%s/proplets/%s/sdf", ts.URL, tc.propletID)
+			res, err := http.Get(reqURL)
+			require.NoError(t, err)
+			defer res.Body.Close()
+
+			assert.Equal(t, tc.wantStatus, res.StatusCode)
+
+			if tc.wantStatus == http.StatusOK {
+				body, err := io.ReadAll(res.Body)
+				require.NoError(t, err)
+
+				var doc sdf.Document
+				require.NoError(t, json.Unmarshal(body, &doc))
+				assert.Equal(t, tc.wantTitle, doc.Info.Title)
+			}
+		})
+	}
+}
--- a/manager/api/transport_test.go
+++ b/manager/api/transport_test.go
@@ -59,11 +59,18 @@ func TestGetPropletSDF(t *testing.T) {
 			wantStatus: http.StatusNotFound,
 		},
 		{
-			desc:       "get SDF with invalid proplet ID returns 500",
+			desc:       "get SDF for non-existent proplet ID returns 404",
 			propletID:  "not-a-valid-uuid",
 			svcDoc:     sdf.Document{},
 			svcErr:     pkgerrors.ErrNotFound,
 			wantStatus: http.StatusNotFound,
+		},
+		{
+			desc:       "get SDF with empty proplet ID returns 400",
+			propletID:  "",
+			svcDoc:     sdf.Document{},
+			svcErr:     nil,
+			wantStatus: http.StatusBadRequest,
 		},
 	}
 
--- a/manager/api/transport_test.go
+++ b/manager/api/transport_test.go
@@ -12,6 +12,7 @@ import (
 	managerapi "github.com/absmach/propeller/manager/api"
 	"github.com/absmach/propeller/manager/mocks"
 	pkgerrors "github.com/absmach/propeller/pkg/errors"
+	"github.com/absmach/propeller/pkg/proplet"
 	"github.com/absmach/propeller/pkg/sdf"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
@@ -59,11 +60,9 @@ func TestGetPropletSDF(t *testing.T) {
 			wantStatus: http.StatusNotFound,
 		},
 		{
-			desc:       "get SDF for non-existent proplet ID returns 404",
+			desc:       "get SDF with invalid proplet ID returns 400",
 			propletID:  "not-a-valid-uuid",
-			svcDoc:     sdf.Document{},
-			svcErr:     pkgerrors.ErrNotFound,
-			wantStatus: http.StatusNotFound,
+			wantStatus: http.StatusBadRequest,
 		},
 		{
 			desc:       "get SDF with empty proplet ID returns 400",
@@ -81,7 +80,9 @@ func TestGetPropletSDF(t *testing.T) {
 			defer ts.Close()
 
 			if tc.propletID != "" {
-				svc.On("GetPropletSDF", mock.Anything, tc.propletID).Return(tc.svcDoc, tc.svcErr)
+				if _, err := uuid.Parse(tc.propletID); err == nil {
+					svc.On("GetPropletSDF", mock.Anything, tc.propletID).Return(tc.svcDoc, tc.svcErr)
+				}
 			}
 
 			reqURL := fmt.Sprintf("%s/proplets/%s/sdf", ts.URL, tc.propletID)
@@ -98,6 +99,68 @@ func TestGetPropletSDF(t *testing.T) {
 				var doc sdf.Document
 				require.NoError(t, json.Unmarshal(body, &doc))
 				assert.Equal(t, tc.wantTitle, doc.Info.Title)
+			}
+		})
+	}
+}
+
+func TestListProplets(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		desc       string
+		query      string
+		svcPage    proplet.PropletPage
+		svcErr     error
+		wantStatus int
+		wantTotal  uint64
+	}{
+		{
+			desc:       "list proplets with defaults",
+			query:      "",
+			svcPage:    proplet.PropletPage{Total: 2, Proplets: []proplet.Proplet{{ID: uuid.NewString(), Name: "p1"}, {ID: uuid.NewString(), Name: "p2"}}},
+			wantStatus: http.StatusOK,
+			wantTotal:  2,
+		},
+		{
+			desc:       "list proplets returns empty page",
+			query:      "",
+			svcPage:    proplet.PropletPage{},
+			wantStatus: http.StatusOK,
+			wantTotal:  0,
+		},
+		{
+			desc:       "list proplets with invalid offset returns 400",
+			query:      "?offset=abc",
+			wantStatus: http.StatusBadRequest,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.desc, func(t *testing.T) {
+			t.Parallel()
+			ts, svc := newServer(t)
+			defer ts.Close()
+
+			if tc.wantStatus != http.StatusBadRequest {
+				svc.On("ListProplets", mock.Anything, uint64(0), mock.AnythingOfType("uint64")).Return(tc.svcPage, tc.svcErr)
+			}
+
+			res, err := http.Get(ts.URL + "/proplets/" + tc.query)
+			require.NoError(t, err)
+			defer res.Body.Close()
+
+			assert.Equal(t, tc.wantStatus, res.StatusCode)
+
+			if tc.wantStatus == http.StatusOK {
+				body, err := io.ReadAll(res.Body)
+				require.NoError(t, err)
+
+				var page struct {
+					Total uint64 `json:"total"`
+				}
+				require.NoError(t, json.Unmarshal(body, &page))
+				assert.Equal(t, tc.wantTotal, page.Total)
 			}
 		})
 	}
--- a/pkg/sdf/proplet.go
+++ b/pkg/sdf/proplet.go
@@ -59,8 +59,7 @@ func PropletDocument(p proplet.Proplet) Document {
 					"stop_task": {
 						Description: "Stop a running task on this proplet",
 						SdfInputData: &DataAfford{
-							Description: "Identifier of the task to stop",
-							Type:        "string",
+							SdfRef: "#/sdfThing/Proplet/sdfData/TaskStop",
 						},
 					},
 				},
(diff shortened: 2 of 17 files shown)


pkg/sdf/sdf.go
```diff
@@ -0,0 +1,74 @@
+package sdf
```
rodneyosodo, line 1: **`minPtr()` placement**: This helper is only called from `proplet.go`. Consider moving it there to keep `sdf.go` a pure type definition file.

pkg/sdf/sdf.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+type EventAfford struct {
+	Description   string      `json:"description,omitempty"`
+	SdfOutputData *DataAfford `json:"sdfOutputData,omitempty"`
+}
+
+type ObjectAfford struct {
+	Description string                    `json:"description,omitempty"`
+	SdfProperty map[string]PropertyAfford `json:"sdfProperty,omitempty"`
```
rodneyosodo: **Unused `ObjectAfford` fields**: `ThingAfford.SdfObject` and `Document.SdfObject` are defined but never populated. Per project convention, remove premature scaffolding — add these when a use case exists.

pkg/sdf/proplet.go
```diff
@@ -0,0 +1,140 @@
...
+							SdfRef: "#/sdfThing/Proplet/sdfData/TaskDispatch",
+						},
+					},
+					"stop_task": {
+						Description: "Stop a running task on this proplet",
+						SdfInputData: &DataAfford{
+							Description: "Identifier of the task to stop",
+							Type:        "string",
+						},
+					},
```
rodneyosodo, line 68: **Inconsistent schema for `stop_task`**: `start_task` references a named `sdfData` entry (`TaskDispatch`), but `stop_task` uses an anonymous inline `DataAfford`. Consider adding a named `TaskStop` schema for consistency:
```go
"stop_task": {
    Description: "Stop a running task on this proplet",
    SdfInputData: &DataAfford{
        SdfRef: "#/sdfThing/Proplet/sdfData/TaskStop",
    },
}
```

rodneyosodo (commented):
Rebase with main

pkg/sdf/proplet.go
```diff
@@ -0,0 +1,140 @@
...
+						Type:        "object",
+						Required:    []string{"id", "name"},
+						Properties: map[string]DataAfford{
+							"id":                {Type: "string"},
+							"name":              {Type: "string"},
+							"image_url":         {Type: "string"},
+							"cli_args":          {Type: "array", Items: &DataAfford{Type: "string"}},
+							"inputs":            {Type: "array", Items: &DataAfford{Type: "string"}},
+							"env":               {Type: "object", AdditionalProperties: &DataAfford{Type: "string"}},
+							"daemon":            {Type: "boolean"},
```
rodneyosodo, line 113: **Open schema for `results`**: No `Type` is set, making this an unconstrained open schema. This is valid SDF but may surprise consumers. If the output is always an array or object, consider constraining it, or at minimum expand the description to set expectations.

pkg/sdf/proplet_test.go
```diff
@@ -0,0 +1,152 @@
+package sdf_test
```
rodneyosodo, line 1: **Missing transport-level tests**: `pkg/sdf` unit tests are solid, but there are no HTTP-level tests for `GET /proplets/{id}/sdf` (e.g. 404 on unknown ID, 200 with correct JSON shape). Other endpoints in `manager/api/` have transport tests — this endpoint should too.

pkg/sdf/sdf.go
JeffMboya: Moved `minPtr()` from `sdf.go` into `proplet.go`. `sdf.go` is now a pure type definition file.

pkg/sdf/sdf.go
JeffMboya: Removed `SdfObject` from both `Document` and `ThingAfford`. No use case exists yet; scaffolding dropped.

pkg/sdf/proplet.go
JeffMboya: Added named `TaskStop` sdfData entry. `stop_task` now references it via `sdfRef`, matching `start_task`'s pattern.

pkg/sdf/proplet.go
JeffMboya: Set `Type: "object"` and added a description clarifying keys and value types depend on the wasm module.

pkg/sdf/proplet_test.go
JeffMboya: Added `manager/api/transport_test.go` with table-driven tests covering 200 OK, 404 unknown ID, and 404 invalid ID.

rodneyosodo (commented):
A few things to fix before merging — one test bug, one dead-code struct, and a few minor nits.

manager/api/transport_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			desc:       "get SDF for existing proplet",
+			propletID:  validID,
+			svcDoc:     validDoc,
+			svcErr:     nil,
+			wantStatus: http.StatusOK,
+			wantTitle:  "test",
+		},
+		{
+			desc:       "get SDF for unknown proplet returns 404",
```
rodneyosodo, line 56: **Bug — test description contradicts `wantStatus`.**

The description says "returns 500" but `wantStatus` is `http.StatusNotFound` (404). These need to match.

Also worth noting: `validate()` on `entityReq` only checks for an empty string — it does not reject non-UUID values. So `"not-a-valid-uuid"` passes validation, reaches the service mock (which returns `ErrNotFound`), and correctly maps to 404. The test behaviour is right but the description is misleading. Consider renaming to `"get SDF for non-existent proplet ID returns 404"` and removing the implication that validation rejects it.

manager/api/transport_test.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+		{
+			desc:       "get SDF with invalid proplet ID returns 500",
+			propletID:  "not-a-valid-uuid",
+			svcDoc:     sdf.Document{},
+			svcErr:     pkgerrors.ErrNotFound,
+			wantStatus: http.StatusNotFound,
+		},
+	}
+
```
rodneyosodo, line 75: **Missing test case: empty proplet ID.**

`validate()` returns `ErrMissingID` when `id == ""`, which should map to a 4xx (likely 400). Add a case with `propletID: ""` and assert it does not reach the service.

pkg/sdf/sdf.go
```diff
@@ -0,0 +1,66 @@
...
```
rodneyosodo: `ObjectAfford` is defined but never used anywhere in this PR or the existing codebase. Either remove it or add a comment explaining when it would be used. Leaving it as dead code will confuse future readers, especially because it is structurally identical to `ThingAfford`.

pkg/sdf/sdf.go
```diff
@@ -0,0 +1,66 @@
...
+	SdfAction   map[string]ActionAfford   `json:"sdfAction,omitempty"`
+	SdfEvent    map[string]EventAfford    `json:"sdfEvent,omitempty"`
+	SdfData     map[string]DataAfford     `json:"sdfData,omitempty"`
+}
+
+type ThingAfford struct {
+	Description string                    `json:"description,omitempty"`
+	SdfProperty map[string]PropertyAfford `json:"sdfProperty,omitempty"`
```
rodneyosodo, line 54: `ThingAfford` and `ObjectAfford` are byte-for-byte identical in structure. If `ObjectAfford` is intentional (e.g. for future nested SDF objects), a type alias (`type ObjectAfford = ThingAfford`) or a shared embedded base struct would prevent them from silently diverging.

pkg/sdf/proplet.go
```diff
@@ -0,0 +1,156 @@
+package sdf
+
+import "github.com/absmach/propeller/pkg/proplet"
+
+func PropletDocument(p proplet.Proplet) Document {
+	return Document{
+		Info: Info{
+			Title:   "Propeller Proplet: " + p.Name + " (" + p.ID + ")",
+			Version: "1.0",
+			License: "Apache-2.0",
```
rodneyosodo, line 14: `Version: "1.0"` is hardcoded and has no tie to the proplet's version or the API version. If the SDF schema changes in future, consumers have no way to detect the old shape from the version field alone. Consider extracting this to a package-level constant (e.g. `const schemaVersion = "1.0"`) so it is easy to find and bump.

pkg/sdk/proplet.go
```diff
@@ -4,12 +4,30 @@ import (
...
+	"github.com/absmach/propeller/pkg/sdf"
 )
 
 const propletsEndpoint = "/proplets"
 
+type Proplet struct {
+	ID        string    `json:"id"`
+	Name      string    `json:"name"`
+	TaskCount uint64    `json:"task_count"`
+	Alive     bool      `json:"alive"`
```
rodneyosodo: `sdk.Proplet` silently drops fields present in `proplet.Proplet` (e.g. `Metadata`, `PassKey`, `Channels`). This is intentional as a stable SDK surface — but it should be documented (a comment on the struct) so the next person editing this file does not assume the two types are interchangeable.

pkg/sdk/proplet.go
```diff
@@ -26,6 +44,52 @@ func (sdk *propSDK) GetPropletAliveHistory(id string, offset, limit uint64) (pro
...
+	}
+
+	var doc sdf.Document
+	if err := json.Unmarshal(body, &doc); err != nil {
+		return sdf.Document{}, err
+	}
+
+	return doc, nil
+}
+
```
rodneyosodo, line 68: Skipping `offset=0` from the query string means an explicit `offset=0` call is indistinguishable from "no offset given". This matches how other SDK methods are written today so it's consistent — just flagging that if the server default ever changes this will silently break callers passing `0` explicitly. `limit=0` has the same issue and is arguably more dangerous (server might interpret it as "unlimited").

manager/api/transport_test.go
JeffMboya: Fixed — description now says 'non-existent proplet ID returns 404'.

manager/api/transport_test.go
JeffMboya: Added empty-ID case; also fixed EncodeError to map ErrValidation → 400.

pkg/sdf/sdf.go
JeffMboya: Replaced with type alias `type ObjectAfford = ThingAfford` — no dead code.

pkg/sdf/sdf.go
JeffMboya: Resolved via type alias; divergence is now structurally impossible.

pkg/sdf/proplet.go
JeffMboya: Extracted to package-level `const schemaVersion = "1.0"`.

pkg/sdk/proplet.go
JeffMboya: Intentional minimal SDK surface; noted, no change per no-comments policy.

pkg/sdk/proplet.go
JeffMboya: Acknowledged — consistent with all other SDK methods today.

manager/api/transport_test.go
```diff
@@ -0,0 +1,104 @@
...
+			svcDoc:     sdf.Document{},
+			svcErr:     pkgerrors.ErrNotFound,
+			wantStatus: http.StatusNotFound,
+		},
+		{
+			desc:       "get SDF for non-existent proplet ID returns 404",
+			propletID:  "not-a-valid-uuid",
+			svcDoc:     sdf.Document{},
+			svcErr:     pkgerrors.ErrNotFound,
+			wantStatus: http.StatusNotFound,
```
rodneyosodo: The expected status here is likely **400**, not 404. `decodeEntityReq` passes the path param through `entityReq.validate()`, which rejects non-UUID values with an `ErrValidation`-wrapped error — hitting the new `case errors.Is(err, apiutil.ErrValidation)` branch added in `pkg/api/api.go`. The mock for `GetPropletSDF` would never be reached in this case. Consider changing `wantStatus` to `http.StatusBadRequest` and removing the `svcErr` field from this test case.

pkg/sdf/sdf.go
```diff
@@ -0,0 +1,60 @@
...
+	SdfAction   map[string]ActionAfford   `json:"sdfAction,omitempty"`
+	SdfEvent    map[string]EventAfford    `json:"sdfEvent,omitempty"`
+	SdfData     map[string]DataAfford     `json:"sdfData,omitempty"`
+}
+
+type ObjectAfford = ThingAfford
```
rodneyosodo: `ObjectAfford` is defined but never used anywhere in this package or the rest of the codebase. Remove it, or add an `sdfObject` field to `Document` and `ThingAfford` (and use it) if you actually need it.

pkg/sdk/proplet.go
```diff
@@ -4,12 +4,30 @@ import (
...
+	"strings"
+	"time"
 
 	"github.com/absmach/propeller/pkg/proplet"
+	"github.com/absmach/propeller/pkg/sdf"
 )
 
 const propletsEndpoint = "/proplets"
 
+type Proplet struct {
```
rodneyosodo, line 20: `sdk.Proplet` and `sdk.PropletPage` duplicate `proplet.Proplet` / `proplet.PropletPage` from `pkg/proplet`. Other SDK methods (e.g. `GetPropletAliveHistory`) reuse domain types directly. If the intent is to expose a stable, separate public shape, add a comment explaining why the duplication is intentional and which fields deliberately differ — otherwise reuse the domain type to avoid silent drift.

pkg/sdk/proplet.go
```diff
@@ -26,6 +44,52 @@ func (sdk *propSDK) GetPropletAliveHistory(id string, offset, limit uint64) (pro
...
+
+	var doc sdf.Document
+	if err := json.Unmarshal(body, &doc); err != nil {
+		return sdf.Document{}, err
+	}
+
+	return doc, nil
+}
+
+func (sdk *propSDK) ListProplets(offset, limit uint64, status string) (PropletPage, error) {
```
rodneyosodo, line 69: `ListProplets` is unrelated to SDF and has no test coverage in this PR. Either move it to a separate PR or add a `TestListProplets` transport test case before merging.

pkg/sdf/proplet.go
```diff
@@ -0,0 +1,158 @@
+package sdf
+
+import "github.com/absmach/propeller/pkg/proplet"
+
+const schemaVersion = "1.0"
```
rodneyosodo, line 7: `schemaVersion = "1.0"` is a magic constant with no link to the proplet's actual version or the SDF spec revision. Document what this version represents (proplet schema version? SDF spec version?), or source it from a build/version constant so it doesn't silently go stale.

pkg/sdf/proplet_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+			for _, key := range tc.wantSdfDataKeys {
+				_, ok := thing.SdfData[key]
+				assert.True(t, ok, "expected sdfData[%q] to exist", key)
+			}
+		})
+	}
+}
+
+func TestPropletDocumentTaskResultStateEnum(t *testing.T) {
```
rodneyosodo, line 82: `TestPropletDocumentTaskResultStateEnum` and `TestPropletDocumentMinimumPointerNotAliased` (line 124) each wrap a single hard-coded case in a table-driven loop. The table adds ceremony without enabling variation — simplify to a plain `t.Run` or inline assertions unless you expect additional cases.

manager/api/transport_test.go
JeffMboya: Fixed. Added UUID validation to entityReq.validate(); non-UUID IDs now return 400 before the service is reached.

pkg/sdf/sdf.go
JeffMboya: Removed. ObjectAfford type alias deleted from sdf.go.

pkg/sdk/proplet.go
JeffMboya: Added comments to both types explaining why the SDK surface intentionally differs from the domain types.

pkg/sdk/proplet.go
JeffMboya: Added TestListProplets transport test covering 200 with results, 200 empty page, and 400 on invalid offset.

pkg/sdf/proplet.go
JeffMboya: Added a comment documenting that schemaVersion tracks the proplet SDF schema definition, not the SDF spec, and should be bumped on breaking changes.

pkg/sdf/proplet_test.go
JeffMboya: Simplified both to inline assertions, removing the single-case table wrapper.

rodneyosodo (approved):

PR #185, opened 2026-04-21
NOISSUE - Add plugin support

JeffMboya:
# What type of PR is this?

Feature — adds a WASM plugin system to the manager.

## What does this do?

- Adds wasmtime-go WASM plugin runtime, replacing wazero.
- Plugin middleware wraps manager with authorize/enrich hooks.
- Rust SDK ships Plugin trait and register_plugin! macro.
- plugin-auth example: authorization and enrichment reference plugin.
- plugin-proxy example rewrites task image URLs to mirrors.
- EnrichResponse extended with ImageURL for registry routing.
- X-User-ID header populates AuthContext.UserID in plugin calls.
- MANAGER_PLUGIN_DIR env var enables plugin directory loading.

No related issue.

Yes — wasm_test.go and middleware plugin_test.go covered.

No — documentation to follow in a separate PR.


JeffMboya:
## plugin-auth example — live test output

The `examples/plugin-auth` plugin demonstrates task ownership enforcement. It runs as a WASM plugin loaded by the manager at startup via `MANAGER_PLUGIN_DIR`.

**How it works:**
- `enrich` (on create): stamps `PROPELLER_CREATED_BY=<user_id>` and `PROPELLER_PLUGIN_AUTHZ=plugin-auth` into the task env
- `authorize` (on start): reads `PROPELLER_CREATED_BY` from task env and denies if the requesting user does not match

**Test run against a live manager (`MANAGER_PLUGIN_DIR=build/plugins`):**

```
# 1. Alice creates a task — enrich stamps her as the owner
$ curl -s -X POST http://localhost:7070/tasks \
    -H "Content-Type: application/json" \
    -H "X-User-Id: alice" \
    -d '{"name":"alice-task"}'

{"id":"f69e5819-...","name":"alice-task","env":{"PROPELLER_CREATED_BY":"alice","PROPELLER_PLUGIN_AUTHZ":"plugin-auth"},...}

# 2. Bob tries to start alice's task — denied by plugin
$ curl -s -X POST http://localhost:7070/tasks/f69e5819-.../start \
    -H "X-User-Id: bob"
{}

# Manager log: "task was created by 'alice', only they can start it"

# 3. Alice starts her own task — plugin allows it
$ curl -s -X POST http://localhost:7070/tasks/f69e5819-.../start \
    -H "X-User-Id: alice"
{}

# Manager log: "Starting task failed" → "no proplet was provided"
# (plugin passed; failure is no proplet registered, not a plugin rejection)
```

**Startup log confirming plugin loaded:**
```
{"level":"INFO","msg":"loaded plugin","name":"plugin-auth","path":"build/plugins/plugin-auth.wasm"}
```

pkg/plugin/wasm.go
```diff
@@ -0,0 +1,281 @@
...
+		alloc:          mod.ExportedFunction(exportAlloc),
+		free:           mod.ExportedFunction(exportFree),
+		auth:           mod.ExportedFunction(exportAuthorize),
+		enrich:         mod.ExportedFunction(exportEnrich),
+		beforeSelect:   mod.ExportedFunction(exportOnBeforeSelect),
+		beforeDispatch: mod.ExportedFunction(exportOnBeforeDispatch),
+		onStart:        mod.ExportedFunction(exportOnStart),
+		onDone:         mod.ExportedFunction(exportOnComplete),
+	}
+
```
rodneyosodo, line 181: **Silent allow-all on missing export — fail-closed instead**

A plugin that exports only `plugin_alloc`/`plugin_free` (but not `authorize`) passes validation and then silently allows every operation. A misconfigured plugin becomes a security bypass with no indication.

```suggestion
	if p.alloc == nil || p.free == nil {
		_ = rt.Close(ctx)

		return nil, fmt.Errorf("plugin %s missing required exports: %s, %s", name, exportAlloc, exportFree)
	}

	if p.auth == nil {
		logger.WarnContext(ctx, "plugin missing authorize export — all requests will be allowed", "plugin", name)
	}
```

Consider making `authorize` a required export alongside alloc/free if you want enforcement to be opt-out rather than opt-in.

rodneyosodo (commented):
Good foundation for a plugin system. I've left inline comments on the key issues found during review. The most important ones to address before merge:

1. **Silent allow-all** when a plugin is missing the `authorize` export
2. **Missing mutex** on `WorkflowCoordinator.SetService`
3. **Fail-open vs fail-closed divergence** between proplet and manager error handling
4. **Two incompatible WASM ABIs** (wazero+JSON for manager, wasmtime+WIT for proplet)

pkg/plugin/wasm.go
```diff
@@ -0,0 +1,281 @@
...
+	if p.alloc == nil || p.free == nil {
+		_ = rt.Close(ctx)
+
+		return nil, fmt.Errorf("plugin %s missing required exports: %s, %s", name, exportAlloc, exportFree)
+	}
+
+	return p, nil
```
rodneyosodo, line 195: **Silent allow-all when `authorize` export is missing**

A plugin that only exports `plugin_alloc`/`plugin_free` passes validation and silently allows every operation (line 136 returns `Allow: true` when `p.auth == nil`). A misconfigured plugin becomes a full security bypass.

Consider logging a loud warning here, or requiring `authorize` alongside the memory exports:

```suggestion
	if p.alloc == nil || p.free == nil {
		_ = rt.Close(ctx)

		return nil, fmt.Errorf("plugin %s missing required exports: %s, %s", name, exportAlloc, exportFree)
	}

	if p.auth == nil {
		logger.WarnContext(ctx, "plugin loaded without authorize export — all requests will be permitted", "plugin", name)
	}

	return p, nil
```

manager/workflow.go
```diff
@@ -25,6 +25,13 @@ func NewWorkflowCoordinator(taskRepo storage.TaskRepository, service Service, lo
 	}
 }
 
+// SetService replaces the Service reference used to start workflow tasks.
+// Call this after all middleware wrapping is complete so workflow-triggered
+// StartTask calls flow through the full plugin middleware chain.
+func (wc *WorkflowCoordinator) SetService(svc Service) {
+	wc.service = svc
```
rodneyosodo, line 36: **Missing mutex — data race vs `cronScheduler.SetService`**

`cronScheduler.SetService` correctly locks `cs.mu` before writing, but this method writes `wc.service` without any synchronisation. Although `SetService` is called before `Subscribe` in `main.go` today, this is a fragile ordering assumption and will be flagged by the race detector if that order ever changes.

```suggestion
func (wc *WorkflowCoordinator) SetService(svc Service) {
	wc.mu.Lock()
	wc.service = svc
	wc.mu.Unlock()
}
```

You will also need to add `mu sync.RWMutex` to the `WorkflowCoordinator` struct and guard reads of `wc.service` with `wc.mu.RLock()` in the same way `cronScheduler` does.

proplet/src/plugin/registry.rs
```diff
@@ -0,0 +1,123 @@
...
+                    allow: false,
+                    reason,
+                }) => {
+                    return Ok(Some(
+                        reason.unwrap_or_else(|| "denied by plugin".to_string()),
+                    ));
+                }
+                Ok(_) => {}
+                Err(e) => {
+                    warn!("Plugin authorize error (treating as allow): {}", e);
```
rodneyosodo: **Fail-open on plugin error contradicts manager behaviour**

The manager-side `authorize` in `plugin.go` returns an error that blocks the operation when a plugin call fails. Here, a plugin error is logged and treated as an allow. The two halves of the system have opposite failure semantics, which will surprise operators running plugins in both places.

```suggestion
                Err(e) => {
                    warn!("Plugin authorize error (failing closed): {}", e);
                    return Ok(Some(format!("plugin error: {e}")));
                }
```

If fail-open is intentional for the proplet side (e.g. because proplets are considered less trusted than the manager), please document the difference explicitly.

manager/middleware/plugin.go
```diff
@@ -0,0 +1,275 @@
...
+
+			return fmt.Errorf("plugin %s authorize: %w", p.Name(), err)
+		}
+		if !resp.Allow {
+			reason := resp.Reason
+			if reason == "" {
+				reason = "denied by plugin"
+			}
+
+			return fmt.Errorf("plugin %s: %s", p.Name(), reason)
```
rodneyosodo: **Error message exposes plugin name to API callers**

The denial reason returned to the HTTP client includes the plugin name (`"plugin auth-plugin: denied"`). This leaks internal infrastructure topology (how many plugins, what they are called) to untrusted clients.

```suggestion
		return fmt.Errorf("%s", reason)
```

The plugin name is already captured in the structured log at line 178 for operators, so it does not need to travel to the caller.

manager/api/transport.go
```diff
@@ -370,6 +373,21 @@ func decodeWorkflowReq(_ context.Context, r *http.Request) (any, error) {
 	return req, nil
 }
 
+func authContextMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		userID := r.Header.Get("X-User-Id")
```
rodneyosodo, line 391: **Header casing: `X-User-Id` vs documented `X-User-ID`**

Go canonicalises HTTP headers so both spellings match on the wire, but the PR description documents `X-User-ID` (uppercase D). Using the non-standard casing here will confuse clients and documentation readers.

```suggestion
		userID := r.Header.Get("X-User-ID")
```

pkg/plugin/wasm.go
```diff
@@ -0,0 +1,281 @@
...
+			break
+		}
+		line := string(bytes.TrimRight(w.buf[:idx], "\r"))
+		w.buf = w.buf[idx+1:]
+		if line != "" {
+			w.logger.Log(context.Background(), w.level, line, "plugin", w.plugin)
+		}
+	}
+
+	return len(p), nil
```
rodneyosodo, line 97: **`slogWriter` silently drops partial lines**

If a plugin writes to stdout/stderr without a trailing newline (e.g. on a panic or crash), the buffered content in `w.buf` is never emitted. Add a `Flush()` method and call it before the runtime is closed:

```go
func (w *slogWriter) Flush() {
	w.mu.Lock()
	defer w.mu.Unlock()
	if len(w.buf) > 0 {
		line := string(bytes.TrimRight(w.buf, "\r\n"))
		if line != "" {
			w.logger.Log(context.Background(), w.level, line, "plugin", w.plugin)
		}
		w.buf = w.buf[:0]
	}
}
```

Then call `stdout.Flush(); stderr.Flush()` in `wasmPlugin.Close` before `p.runtime.Close(ctx)`.

Makefile
```diff
@@ -201,8 +201,18 @@ hal-test:
 	cp examples/hal-test/target/wasm32-wasip1/release/hal-test.wasm build/hal-test.wasm
 
 attestation-test:
-	cd examples/attestation-test && cargo build --target wasm32-wasip2 --release
-	cp examples/attestation-test/target/wasm32-wasip2/release/attestation_test.wasm build/attestation-test.wasm
+	cd examples/attestation-test && cargo build --release
```
rodneyosodo: **Unrelated change: `attestation-test` silently switches from `wasm32-wasip2` to `wasm32-wasip1`**

This is a behaviour change (different WASM target) that is not mentioned in the PR description and has nothing to do with the plugin system. If this is a bug fix it should be in a separate commit with an explanation; if it is accidental it should be reverted.

```suggestion
attestation-test:
	cd examples/attestation-test && cargo build --target wasm32-wasip2 --release
	cp examples/attestation-test/target/wasm32-wasip2/release/attestation_test.wasm build/attestation-test.wasm
```

pkg/plugin/registry.go
```diff
@@ -0,0 +1,82 @@
...
+type staticRegistry struct {
+	plugins []Plugin
+	logger  *slog.Logger
+}
+
+func LoadDirectory(ctx context.Context, dir string, logger *slog.Logger) (Registry, error) {
+	r := &staticRegistry{logger: logger}
+
+	if dir == "" {
+		return r, nil
```
rodneyosodo, line 22: **`nil` registry guard in `service.go` is dead code**

`LoadDirectory` always returns a non-nil `*staticRegistry` (even for an empty dir). The `if svc.plugins == nil` guards in `service.go` are therefore never true in production and create a false sense that a nil registry is a valid state.

Consider one of two options:

1. **Keep `nil` as a valid no-op state** — return `nil, nil` here instead of a zero-value registry:
```suggestion
	if dir == "" {
		return nil, nil
	}
```
…and update callers to handle `nil` (which they already do).

2. **Remove the nil guards in `service.go`** and always pass the result of `LoadDirectory` (which is always non-nil with an empty plugin list when `MANAGER_PLUGIN_DIR` is unset).

pkg/plugin/wasm.go
JeffMboya: Added WarnContext when authorize export is absent.

pkg/plugin/wasm.go
JeffMboya: Added WarnContext when authorize export is absent.

manager/workflow.go
JeffMboya: Added sync.RWMutex; SetService writes, reads use RLock.

proplet/src/plugin/registry.rs
JeffMboya: Changed to fail-closed; returns error instead of allowing.

manager/middleware/plugin.go
JeffMboya: Stripped plugin name; returns only the denial reason.

manager/api/transport.go
JeffMboya: Fixed to X-User-ID throughout.

pkg/plugin/wasm.go
JeffMboya: Added Flush(); called in Close before runtime.Close.

Makefile
JeffMboya: Reverted to wasm32-wasip2 target.

pkg/plugin/registry.go
JeffMboya: Returns nil,nil when dir empty; callers guard against nil.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,113 @@
...
+	PropletID string            `json:"proplet_id,omitempty"`
+	Priority  int               `json:"priority,omitempty"`
+	Daemon    bool              `json:"daemon,omitempty"`
+	Encrypted bool              `json:"encrypted,omitempty"`
+	CreatedAt time.Time         `json:"created_at,omitzero"`
+}
+
+type AuthContext struct {
+	UserID string `json:"user_id,omitempty"`
+	Token  string `json:"token,omitempty"`
```
rodneyosodo: **Security: Bearer token passed to untrusted WASM plugins**

`Token` is the caller's `Authorization` Bearer token, serialised as JSON and handed to every loaded plugin. A compromised or malicious plugin can read, log, or exfiltrate it. Authorization decisions need only `UserID` and `Action`. Consider removing `Token` from `AuthContext`, or omitting it from serialisation by default and only including it when the plugin has been explicitly granted access.

pkg/plugin/wasm.go
```diff
@@ -0,0 +1,304 @@
...
+	exportAlloc            = "plugin_alloc"
+	exportFree             = "plugin_free"
+	exportAuthorize        = "authorize"
+	exportEnrich           = "enrich_task"
+	exportOnBeforeSelect   = "on_before_proplet_select"
+	exportOnBeforeDispatch = "on_before_dispatch"
+	exportOnStart          = "on_task_start"
+	exportOnComplete       = "on_task_complete"
+
+	pluginCallTimeout = 5 * time.Second
```
rodneyosodo: **Hardcoded 5-second timeout blocks the entire task start hot path**

Every `StartTask` holds the per-plugin mutex (`p.mu`) for up to 5 seconds waiting for `Authorize`. A slow or hung plugin queues every concurrent start behind it. Consider making this configurable (e.g. `MANAGER_PLUGIN_CALL_TIMEOUT`) and defaulting to something shorter (e.g. 500 ms).

manager/middleware/plugin_test.go
```diff
@@ -0,0 +1,363 @@
...
+			pl.On("Authorize", mock.Anything, mock.Anything).Return(tc.authResp, nil)
+
+			if !tc.wantErr {
+				svc.On("DeleteTask", mock.Anything, "task-1").Return(nil)
+			}
+
+			err := mw.DeleteTask(context.Background(), "task-1")
+			if tc.wantErr {
+				require.Error(t, err)
+				assert.Equal(t, "plugin mock-plugin: no", err.Error())
```
rodneyosodo: **Wrong expected error string — test will fail**

The deny path in `authorize` returns `fmt.Errorf("%s", reason)`, which produces the string `"no"`. This assertion expects `"plugin mock-plugin: no"`, which is only produced by the hard-error path (`fmt.Errorf("plugin %s authorize: %w", ...)`). Change to:
```go
assert.Equal(t, "no", err.Error())
```

manager/middleware/plugin.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (pm *pluginMiddleware) CreateWorkflow(ctx context.Context, tasks []task.Task) ([]task.Task, error) {
+	for i := range tasks {
+		if err := pm.authorize(ctx, plugin.ActionCreate, toTaskInfo(tasks[i])); err != nil {
+			return nil, err
+		}
+	}
+
+	return pm.Service.CreateWorkflow(ctx, tasks)
```
rodneyosodo, line 110: **`CreateWorkflow` authorizes but never enriches**

`CreateTask` calls both `authorize` and `enrich`/`applyEnrichment`, so plugin-injected env vars (e.g. `PROPELLER_CREATED_BY` from the reference plugin) are applied before the task is persisted. `CreateWorkflow` and `CreateJob` only call `authorize`, so those same vars are never added. Workflow tasks will be missing the enrichment that individually-created tasks receive.

manager/service.go
```diff
@@ -1819,6 +1835,141 @@ func (svc *service) publishStart(ctx context.Context, t task.Task, propletID str
 	return svc.pubsub.Publish(ctx, topic, payload)
 }
 
+func (svc *service) runOnBeforePropletSelect(ctx context.Context, t task.Task) (plugin.PropletSelectResponse, error) {
+	merged := plugin.PropletSelectResponse{Allow: true}
+	req := plugin.PropletSelectRequest{Context: plugin.AuthFromContext(ctx), Task: toPluginTaskInfo(t)}
+
+	for _, p := range svc.plugins.List() {
```
rodneyosodo, line 1845: **Nil `plugin.Registry` panics here**

All test helpers pass `nil` as `plugins` (e.g. `service_proplet_test.go`, `service_workflow_test.go`, `shutdown_test.go`). Calling `.List()` on a nil interface panics. `runOnBeforeDispatch` and `notifyTaskComplete` have the same issue. Add a nil guard:
```go
if svc.plugins == nil {
    return plugin.PropletSelectResponse{Allow: true}, nil
}
```

manager/service.go
```diff
@@ -1819,6 +1835,141 @@ func (svc *service) publishStart(ctx context.Context, t task.Task, propletID str
...
+		svc.wg.Go(func() {
+			if err := p.OnTaskComplete(detached, evt); err != nil {
+				svc.logger.WarnContext(detached, "plugin on_task_complete failed", "plugin", p.Name(), "error", err)
+			}
+		})
+	}
+}
+
+func toPluginTaskInfo(t task.Task) plugin.TaskInfo {
```
rodneyosodo: **Duplicate helper: `toPluginTaskInfo` is identical to `toTaskInfo` in `manager/middleware/plugin.go:245`**

Both functions convert `task.Task` → `plugin.TaskInfo` field-for-field. Consider exporting a single `plugin.FromTask(t task.Task) plugin.TaskInfo` utility or moving the canonical implementation to `pkg/plugin` to avoid the two copies drifting.

go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	github.com/jtolds/gls v4.20.0+incompatible // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/stretchr/objx v0.5.3 // indirect
+	github.com/tetratelabs/wazero v1.11.0 // indirect
```
rodneyosodo: **`wazero` should be a direct dependency, not `// indirect`**

`pkg/plugin/wasm.go` imports `github.com/tetratelabs/wazero` directly. Run `go mod tidy` to promote it to a direct dependency.

rodneyosodo (commented):
See individual comments for details.

pkg/plugin/plugin.go
JeffMboya: Stripped Token from JSON serialization with json:"-" — plugins never see the bearer token.

pkg/plugin/wasm.go
JeffMboya: Removed fixed timeout; epoch-based interruption now honours caller context deadline.

manager/middleware/plugin_test.go
JeffMboya: Fixed — test now asserts "no", matching the stripped denial reason.

manager/middleware/plugin.go
JeffMboya: Added enrich step to CreateWorkflow and CreateJob, matching CreateTask.

manager/service.go
JeffMboya: Added nil guards in runOnBeforePropletSelect, runOnBeforeDispatch, notifyTaskComplete.

manager/service.go
JeffMboya: Consolidated into plugin.NewTaskInfo; both private helpers removed.

go.mod
JeffMboya: Switched from wazero to wasmtime-go; now a direct dependency.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/rubenv/sql-migrate v1.8.1
 	github.com/spf13/cobra v1.10.2
+	github.com/bytecodealliance/wasmtime-go v1.0.0
```
rodneyosodo: `wasmtime-go v1.0.0` was released in mid-2022; the library is now in the `v44+` range. Older versions miss security patches, API improvements, and WASI preview-1 compatibility fixes. Please upgrade to the latest stable release.

```suggestion
	github.com/bytecodealliance/wasmtime-go v44.0.0
```

go.sum
```diff
@@ -261,6 +263,8 @@ github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UV
 github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu7U=
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
+github.com/tetratelabs/wazero v1.11.0 h1:+gKemEuKCTevU4d7ZTzlsvgd1uaToIDtlQlmNbwqYhA=
```
rodneyosodo: The PR description says wasmtime-go **replaces** wazero, but `wazero v1.11.0` is still present in `go.sum`. If it's genuinely gone, `go mod tidy` should have removed it. If it's still a transitive dependency, please call that out — otherwise this is a stale entry that inflates the build graph.

manager/api/transport.go
```diff
@@ -273,6 +275,7 @@ func decodeTaskReq(_ context.Context, r *http.Request) (any, error) {
 
 func decodeUploadTaskFileReq(_ context.Context, r *http.Request) (any, error) {
 	var req taskReq
+	r.Body = http.MaxBytesReader(nil, r.Body, maxFileSize)
```
rodneyosodo: Passing `nil` as the `ResponseWriter` to `MaxBytesReader` means it cannot set `Connection: close` or write a 413 on overflow — the error is returned but the client receives no hint about the limit. Because the go-kit decoder signature `func(_ context.Context, r *http.Request)` has no access to `w`, the cleaner fix is a Chi middleware applied before the route:

```go
// in MakeHandler, before route registration:
mux.Use(func(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        r.Body = http.MaxBytesReader(w, r.Body, maxFileSize)
        next.ServeHTTP(w, r)
    })
})
```

Then remove the `MaxBytesReader` call from `decodeUploadTaskFileReq`.

manager/api/transport.go
rodneyosodo, line 391: **Security — unauthenticated trust of `X-User-Id`.**

Any HTTP client can forge this header; there is no validation or signature. If plugin-auth logic relies on `UserID` to enforce ownership checks (e.g. "only the creator can start this task"), forging the header trivially bypasses that gate.

Options:
1. **Trusted-proxy only**: document that the manager must sit behind an authenticating reverse proxy that strips and re-sets `X-User-Id`, and reject requests that arrive with the header on the public interface.
2. **Token validation**: validate the `Authorization` bearer token and derive `UserID` from it server-side rather than trusting the client-supplied header.
3. **Advisory only**: document clearly that `UserID` is advisory (for logging/enrichment) and not an authorization boundary — and remove any plugin logic that uses it to block actions.

At minimum, add a comment here explaining the trust model so future contributors don't assume this is authenticated.

manager/middleware/plugin.go
```diff
@@ -0,0 +1,272 @@
...
+
+			return fmt.Errorf("plugin %s authorize: %w", p.Name(), err)
+		}
+		if !resp.Allow {
+			reason := resp.Reason
+			if reason == "" {
+				reason = "denied by plugin"
+			}
+
+			return fmt.Errorf("%s", reason)
```
rodneyosodo: `fmt.Errorf("%s", reason)` adds no error wrapping and the `%s` verb is redundant. Use `errors.New` instead — it's cleaner and avoids a `go vet` / `staticcheck` warning about non-constant format strings.

```suggestion
			return errors.New(reason)
```

manager/middleware/plugin.go
```diff
@@ -0,0 +1,272 @@
...
+			return "", nil, err
+		}
+		enriched := pm.enrich(ctx, info)
+		applyEnrichment(&tasks[i], enriched)
+	}
+
+	return pm.Service.CreateJob(ctx, name, tasks, executionMode)
+}
+
+func (pm *pluginMiddleware) StartJob(ctx context.Context, jobID string) error {
```
rodneyosodo, line 126: `StartJob` authorizes every task but **does not enrich** them, which is inconsistent with `CreateJob` (lines 108–122) that does both. Plugins that inject env vars via `Enrich` (e.g. `PROPELLER_CREATED_BY` from the reference plugin) will have stamped those vars at create-time but they won't be re-applied at start-time for job tasks.

If enrichment at start is intentional no-op, add a comment explaining why. If it's an oversight, apply the same pattern as `CreateJob`:

```suggestion
func (pm *pluginMiddleware) StartJob(ctx context.Context, jobID string) error {
	tasks, err := pm.GetJob(ctx, jobID)
	if err != nil {
		return err
	}

	for i := range tasks {
		info := plugin.NewTaskInfo(tasks[i])
		if err := pm.authorize(ctx, plugin.ActionStart, info); err != nil {
			return err
		}
		enriched := pm.enrich(ctx, info)
		applyEnrichment(&tasks[i], enriched)
	}

	return pm.Service.StartJob(ctx, jobID)
}
```

manager/service.go
```diff
@@ -1819,6 +1835,138 @@ func (svc *service) publishStart(ctx context.Context, t task.Task, propletID str
...
+		}
+		if !resp.Allow {
+			reason := resp.Reason
+			if reason == "" {
+				reason = "proplet selection blocked by plugin"
+			}
+
+			return plugin.PropletSelectResponse{}, errors.New(reason)
+		}
+		merged.RequiredTags = append(merged.RequiredTags, resp.RequiredTags...)
```
rodneyosodo: Tags from multiple plugins are appended without deduplication. If two plugins both require the same tag, the scheduler sees it twice. The `Inputs` merge loop (in `applyEnrichment`) already deduplicates with `slices.Contains` — apply the same pattern here for consistency.

```suggestion
		for _, tag := range resp.RequiredTags {
			if !slices.Contains(merged.RequiredTags, tag) {
				merged.RequiredTags = append(merged.RequiredTags, tag)
			}
		}
```

.github/dependabot.yml
```diff
@@ -33,6 +33,7 @@ updates:
       - "examples/attestation-test"
       - "examples/http-client"
       - "examples/http-server"
+      - "examples/plugin-auth"
```
rodneyosodo, line 36: The PR adds `examples/plugin-proxy` and `examples/proplet-plugin-example` as new Cargo workspaces, but only `examples/plugin-auth` was added to dependabot. Please add the other two so their dependencies get automated update PRs.

```suggestion
      - "examples/plugin-auth"
      - "examples/plugin-proxy"
      - "examples/proplet-plugin-example"
```

crates/propeller-plugin-sdk/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        #[no_mangle]
+        pub extern "C" fn plugin_alloc(len: u32) -> u32 {
+            $crate::__private::alloc_impl(len)
+        }
+
+        #[no_mangle]
+        pub unsafe extern "C" fn plugin_free(ptr: u32, len: u32) {
+            if ptr != 0 {
+                let _ = ::std::vec::Vec::from_raw_parts(ptr as *mut u8, 0, len as usize);
```
rodneyosodo, line 240: The correctness of `plugin_free` depends on a hidden invariant: `alloc_impl(len)` creates `Vec::with_capacity(len)` (length 0, capacity `len`), so reconstructing with `from_raw_parts(ptr, 0, len)` correctly deallocates `len` bytes. This works, but the invariant is invisible at the call site. A short comment prevents future confusion:

```suggestion
            if ptr != 0 {
                // Safety: ptr was returned by alloc_impl(len), which allocates
                // Vec::with_capacity(len). Reconstruct with the same capacity so
                // the allocator frees the correct number of bytes.
                let _ = ::std::vec::Vec::from_raw_parts(ptr as *mut u8, 0, len as usize);
            }
```

proplet/src/plugin/registry.rs
```diff
@@ -0,0 +1,124 @@
...
+            match plugin.enrich(task.clone()) {
+                Ok(EnrichResponse { env: additions }) => env.extend(additions),
+                Err(e) => warn!("Plugin enrich error (skipping): {}", e),
+            }
+        }
+        Ok(env)
+    }
+
+    /// Fire-and-forget: notifies all plugins that a task started.
+    pub fn notify_task_start(registry: Arc<Self>, task: TaskInfo) {
```
rodneyosodo: `spawn_blocking` returns a `JoinHandle` that is immediately dropped here, so there is no way to:
- observe panics inside the blocking task
- await completion during graceful shutdown
- cancel an in-flight notification

Consider storing the handles or using a `tokio::task::JoinSet` that the caller can drain at shutdown:

```rust
pub fn notify_task_start(registry: Arc<Self>, task: TaskInfo) -> tokio::task::JoinHandle<()> {
    tokio::task::spawn_blocking(move || {
        for plugin in &registry.plugins {
            plugin.on_task_start(task.clone());
        }
    })
}
```

The same applies to `notify_task_complete`.

rodneyosodo (changes requested):
Good overall architecture — the middleware chain wiring, epoch-based WASM timeout, `context.WithoutCancel` for fire-and-forget notifications, and the bounds-checked `UnsafeData` reads are all done correctly. A few items below need attention before merge.

go.mod
JeffMboya: Upgraded to github.com/bytecodealliance/wasmtime-go/v44 v44.0.0 and ran go mod tidy.

go.sum
JeffMboya: go mod tidy removed wazero from go.sum — it was a stale transitive entry from the wazero era.

manager/api/transport.go
JeffMboya: Moved MaxBytesReader into a Chi middleware in MakeHandler with ResponseWriter available; removed from decodeUploadTaskFileReq.

manager/api/transport.go
JeffMboya: Added doc comment to authContextMiddleware explaining the trust model and proxy requirement.

manager/middleware/plugin.go
JeffMboya: Replaced with errors.New(reason); errors import added.

manager/middleware/plugin.go
JeffMboya: Added enrich call to StartJob with comment explaining enrichment is intentional no-op at start time (env vars persisted at create).

manager/service.go
JeffMboya: Deduplicated RequiredTags using slices.Contains, matching the Inputs merge pattern.

.github/dependabot.yml
JeffMboya: Added examples/plugin-proxy and examples/proplet-plugin-example to the Cargo dependabot entries.

crates/propeller-plugin-sdk/src/lib.rs
JeffMboya: Added Safety comment documenting the alloc_impl invariant that makes the capacity reconstruction correct.

proplet/src/plugin/registry.rs
JeffMboya: Both functions now return JoinHandle<()>; call sites use let _ = ... to acknowledge the handle explicitly.

rodneyosodo (approved):
LGTM

Issue #189, opened 2026-04-22
Broadcast feature is incomplete


rodneyosodo:
## Description

The broadcast feature appears incomplete - the manager publishes broadcast tasks but the proplet cannot receive or process them.

### Issue 1: Broadcast tasks cannot be stopped

The `StopTask` function fails for broadcast tasks. When a broadcast task is stopped:
- `svc.taskPropletRepo.Get(ctx, taskID)` returns an error because no proplet is assigned to broadcast tasks
- This error is not handled, causing the stop operation to fail

Broadcast tasks cannot be stopped once started.

### Issue 2: Broadcast field not included in start payload

The `publishStart` function doesn't include the `broadcast` field in the message payload. The proplet needs to know whether it's a broadcast task to skip the proplet_id check (proplet/src/service.rs:436-440).

Proplets won't correctly filter broadcast messages.

### Issue 3: Proplet doesn't read broadcast from StartRequest

The `StartRequest` struct has no `broadcast` field. The `validate` function exists but the struct doesn't include the broadcast field in the deserialized JSON.

Even if the manager sends broadcast, the proplet won't receive or process it.

## Additional Notes

These gaps mean the broadcast feature won't work as intended. The manager properly publishes broadcast tasks and the field is added to the task model and DB, but the proplet-side processing is missing.

PR #190, opened 2026-04-22
NOISSUE - FIx Broadcast Support

rodneyosodo:
# What type of PR is this?

This is a bug fix — it fixes broadcast task support so tasks are dispatched to all proplets correctly, and ensures non-broadcast tasks are only executed by the targeted proplet.

## What does this do?

- Fixes `StartTask`/`StopTask` in the manager to handle broadcast tasks without requiring a pinned proplet
- Fixes `publishStart` to send `null` instead of `""` for `proplet_id` on broadcast tasks
- Adds validation in `CreateTask` rejecting tasks with both `broadcast: true` and a non-empty `proplet_id`
- Adds proplet-side filtering so non-broadcast start commands are ignored if `proplet_id` does not match the receiving proplet
- Updates Rust `StartRequest` test structs to include the new `broadcast` field

#189 

Updated existing Rust unit tests to include the new `broadcast` field in `StartRequest` struct literals.

No additional documentation required.

### Notes

The proplet-side `proplet_id` check also fixes a pre-existing bug where all proplets would execute every non-broadcast task regardless of the targeted proplet ID.

### DIFF
--- a/manager/service.go
+++ b/manager/service.go
@@ -156,6 +156,10 @@ func (svc *service) DeleteProplet(ctx context.Context, propletID string) error {
 }
 
 func (svc *service) CreateTask(ctx context.Context, t task.Task) (task.Task, error) {
+	if t.Broadcast && t.PropletID != "" {
+		return task.Task{}, errors.New("proplet_id must not be set when broadcast is true")
+	}
+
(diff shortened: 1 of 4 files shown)


JeffMboya (approved):
LGTM

PR #191, opened 2026-04-22
NOISSUE - Optimize task handling and fix dependency validation

rodneyosodo:
# What type of PR is this?
This is a bug fix and optimisation because it addresses several correctness and security issues found during a code review.

## What does this do?
  - Fix double-close panic in CronScheduler.Stop(): Guards close(stopChan) with sync.Once so calling Stop() more than once no longer panics.
  - Fix unbounded recursion in round-robin scheduler: Replaces the recursive SelectProplet call (which could stack-overflow with many dead proplets) with an iterative loop.
  - Optimise TopologicalSort from O(n²) to O(n+e): Builds a reverse adjacency map (dependents) up front so the BFS inner step is O(1) per edge instead of scanning all tasks on every iteration.
  - Fix TopologicalSort missing dependency validation: ValidateDependenciesExist is now called before ValidateDAG, ensuring tasks with references to non-existent dependencies are caught early.
  - Fix world-readable config file permissions: Changes filePermission from 0o644 to 0o600 so provisioned config files (which contain MQTT credentials) are not readable by other users on the system.
  - Clarify RunIfFailure semantics: Adds a comment explaining the intentional asymmetry — success requires all dependencies to complete, failure triggers if any dependency fails.
  - Batch job listing: Refactors ListJobs to fetch all tasks once and build a job map, rather than issuing a separate ListByJobID query per stored job.
  - Makefile default goal: Sets .DEFAULT_GOAL := all so a bare make builds everything.

N/A

No user-facing features were added. The RunIfFailure behaviour is documented inline with a code comment.

The round-robin fix also removes the p.TaskCount += 1 increment that was present in the old path but operated on a copy of the struct (the increment was already a no-op).

### Notes
N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,8 @@ define docker_push
 		done
 endef
 
+.DEFAULT_GOAL := all
+
 $(SERVICES):
 	$(call compile_service,$(@))
 
--- a/cli/provision.go
+++ b/cli/provision.go
@@ -25,7 +25,7 @@ var (
 	fileName = "config.toml"
 )
 
-const filePermission = 0o644
+const filePermission = 0o600
 
 func SetSuperMQSDK(sdk smqSDK.SDK) {
 	smqsdk = sdk
--- a/manager/cron.go
+++ b/manager/cron.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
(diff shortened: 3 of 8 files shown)


JeffMboya (approved):
LGTM

PR #212, opened 2026-05-29
NOISSUE - Add HAL host and WIT interfaces with linker support

rodneyosodo:
# What type of PR is this?


rodneyosodo:
Already implemented

PR #219, opened 2026-06-01
NOISSUE - Use attestation-agent config and add Wasmtime cache

rodneyosodo:
# What type of PR is this?

### DIFF
--- a/hal/ubuntu/qemu.sh
+++ b/hal/ubuntu/qemu.sh
@@ -151,15 +151,23 @@ write_files:
       PROPLET_ENABLE_MONITORING=true
       PROPLET_MANAGER_K8S_NAMESPACE=default
       PROPLET_KBS_URI=KBS_URI_PLACEHOLDER
-      PROPLET_AA_CONFIG_PATH=/etc/default/proplet.toml
+      PROPLET_AA_CONFIG_PATH=/etc/attestation-agent.conf
       PROPLET_LAYER_STORE_PATH=/tmp/proplet/layers
     permissions: '0644'
 
-  - path: /etc/default/proplet.toml
+  - path: /etc/attestation-agent.conf
     content: |
       [token_configs]
-      [token_configs.coco_kbs]
+
+      [token_configs.kbs]
       url = "KBS_URI_PLACEHOLDER"
+
+      [eventlog_config]
+      init_pcr = 17
+      enable_eventlog = false
+
+      [log]
+      level = "info"
     permissions: '0644'
 
   - path: /etc/default/attestation-agent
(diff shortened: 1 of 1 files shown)


hal/ubuntu/qemu.sh
```diff
@@ -151,15 +151,23 @@ write_files:
...
       PROPLET_LAYER_STORE_PATH=/tmp/proplet/layers
     permissions: '0644'
 
-  - path: /etc/default/proplet.toml
+  - path: /etc/attestation-agent.conf
     content: |
       [token_configs]
-      [token_configs.coco_kbs]
+
+      [token_configs.kbs]
```
JeffMboya, line 162: `kbs` renamed from `coco_kbs`; AA version must support this key.

hal/ubuntu/qemu.sh
```diff
@@ -259,7 +267,9 @@ write_files:
       Type=simple
       EnvironmentFile=/etc/default/proplet
       Environment=WASMTIME_HOME=/var/lib/proplet
+      Environment=WASMTIME_CACHE_DIR=/var/cache/wasmtime
```
JeffMboya, line 270: `WASMTIME_CACHE_DIR` is not recognized by Wasmtime; use `WASMTIME_CACHE_CONFIG`.

hal/ubuntu/qemu.sh
rodneyosodo: Yes. It does

hal/ubuntu/qemu.sh
rodneyosodo: Wasmtime has neither env variables. It is just unix style https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s05.html

JeffMboya (approved):
LGTM

hal/ubuntu/qemu.sh
```diff
@@ -271,7 +281,7 @@ write_files:
       PrivateTmp=true
       ProtectSystem=strict
       ProtectHome=true
-      ReadWritePaths=/var/lib/proplet /tmp
+      ReadWritePaths=/var/lib/proplet /var/cache/wasmtime /tmp
```
drasko, line 284: Why are we hard-coding for only one Wasm runtime? Who is guaranteeing that only Wasmtime will be used? Are we hard-coding this on purpose?

drasko (changes requested):

hal/ubuntu/qemu.sh
rodneyosodo: We are shipping the qemu config with the default wasmtime installed

PR #226, opened 2026-06-03
NOISSUE - Implement `wasi-usb`

Michielvk:
# What type of PR is this?

Feature: allows the Proplet to interact with USB peripherals. E.g., workload needs access to a USB sensor.

## What does this do?

Add support for USB using `wasmtime-wasi-usb` into the Wasmtime embedded runtime.

ELASTIC Demonstrator 1: WasmHAL-HW

N/A

Planned once changes are approved to be upstreamed.

### Notes

- USB transfers (often) require async calls/callbacks. Our bindings thus require async. I saw that there was a distinction in the current proplet between a sync and async `Engine` . Is this something historical, or is this a deliberate design decision? The reason I ask is that maintaining two engines would result in a lot of code duplication for usb support.
- I currently opted for a single `async` `Engine` for everything. This removes the need to maintain two Engines, since the async one was only used for `wasi-http`. An async Engine is perfectly capable of running sync components. Instead of moving the sync execution to a blocking thread, we just run it in the async Engine and let Wasmtime handle the blocking call. A long blocking call will block the tokio worker thread (I think, not 100% sure if/how Wasmtime handles this), so a malicious Wasm binary might cause starvation if all worker threads are blocked (`spawn_blocking` actually has the same flaw, but less risky). We might need to implement https://docs.wasmtime.dev/examples-interrupting-wasm.html, but that comes at a performance hit. So my question is: do you think there is a better solution to what I did currently, or is this fine for now?
- Will switching to an async Engine for everything have other bad/unwanted implications that I didn't think of?

### DIFF
--- a/docker/Dockerfile.proplet
+++ b/docker/Dockerfile.proplet
@@ -47,7 +47,7 @@ LABEL org.opencontainers.image.licenses=Apache-2.0
 LABEL org.opencontainers.image.title="Propeller Proplet (Rust)"
 LABEL org.opencontainers.image.description="Propeller edge runtime for WebAssembly modules"
 
-ARG WASMTIME_VERSION=42.0.1
+ARG WASMTIME_VERSION=44.0.0
 ARG WASMTIME_SHA256_AMD64=dd5253f3cb521bb094f9951c3d2c45c746b31e5723b07ce56f162ec9bab44d59
 ARG WASMTIME_SHA256_ARM64=fa9b7e09f49f75c17acf2c018a4286cdbeffb4c1f3ee9e72c48b6a42c1deceda
 ARG WASMTIME_SHA256_RISCV64=04fd303b53315650a6cb7230d5a606c4d72187bee439a7ba269c5b2dade8712a
--- a/docker/Dockerfile.proplet
+++ b/docker/Dockerfile.proplet
@@ -48,9 +48,9 @@ LABEL org.opencontainers.image.title="Propeller Proplet (Rust)"
 LABEL org.opencontainers.image.description="Propeller edge runtime for WebAssembly modules"
 
 ARG WASMTIME_VERSION=44.0.0
-ARG WASMTIME_SHA256_AMD64=dd5253f3cb521bb094f9951c3d2c45c746b31e5723b07ce56f162ec9bab44d59
-ARG WASMTIME_SHA256_ARM64=fa9b7e09f49f75c17acf2c018a4286cdbeffb4c1f3ee9e72c48b6a42c1deceda
-ARG WASMTIME_SHA256_RISCV64=04fd303b53315650a6cb7230d5a606c4d72187bee439a7ba269c5b2dade8712a
+ARG WASMTIME_SHA256_AMD64=52eba06fe9f4364aa6164a4a3eafb2ca692ba9a756cbe8137b5574871f8cbfc8
+ARG WASMTIME_SHA256_ARM64=294cae921fb88cbbcb60a914eaaaf313df3249d718609afb5804186b3f1912f5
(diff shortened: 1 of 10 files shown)


Michielvk:
@rodneyosodo

docker/Dockerfile.proplet
```diff
@@ -47,7 +47,7 @@ LABEL org.opencontainers.image.licenses=Apache-2.0
 LABEL org.opencontainers.image.title="Propeller Proplet (Rust)"
 LABEL org.opencontainers.image.description="Propeller edge runtime for WebAssembly modules"
 
-ARG WASMTIME_VERSION=42.0.1
+ARG WASMTIME_VERSION=44.0.0
 ARG WASMTIME_SHA256_AMD64=dd5253f3cb521bb094f9951c3d2c45c746b31e5723b07ce56f162ec9bab44d59
```
JeffMboya: SHA256 hashes not updated for v44; builds will fail.

proplet/Cargo.toml
```diff
@@ -63,6 +63,9 @@ elastic-tee-hal = { git = "https://github.com/elasticproject-eu/wasmhal", rev =
 ] }
 wasm-wave = "0.246.2"
 
+# ELASTIC WasmHAL-Hardware: wasi-usb
+wasmtime-wasi-usb = { git = "https://github.com/idlab-discover/wasmtime-wasi-usb", rev = "bc7c92d" }
```
JeffMboya: Git dep pinned to single commit; not published to crates.io.

proplet/src/runtime/wasmtime_runtime.rs
```diff
@@ -385,6 +393,7 @@ impl WasmtimeRuntime {
         let store_data = StoreData {
             wasi,
             http: WasiHttpCtx::new(),
+            usb: WasiUsbCtx::new(false),
```
JeffMboya, line 396: Hardcoded `false`; USB disabled even when `usb_enabled` is true.

Michielvk: `wasmtime-wasi-usb` has support for USB hotplug support. Because this is currently a `wasip2` interface, we added a toggle to disable it to disable the hotplug poll-loop. The goal is to later either remove this toggle in `wasip3`, or to use a cargo feature instead.

We don't think hotplug support makes sense in Propeller, since most of the use-cases are ephemeral tasks, that's why we disabled it here. It would be weird to support hotplug and wait for someone to plug-in a USB device on an edge node.

```rust
impl WasiUsbCtx {
    pub fn new(hotplug_enabled: bool) -> Self {
        Self {
            libusb_ctx: None,
            event_loop_flag: None,
            event_thread: None,
            hotplug_enabled,
            hotplug_handle: None,
        }
    }
}
```

https://github.com/idlab-discover/wasmtime-wasi-usb/blob/2717ebe5ed04aa2d760b8bf087f11b154ce02d3d/src/usb.rs#L27-L37

docker/Dockerfile.proplet
Michielvk: Fixed in 856eccc4cf97ce6e6f9d7cb4c18bb7e85ce5c7e2

proplet/Cargo.toml
Michielvk: Fixed in 1eec3a95806c183c7504f64cddc6df663bd0f260.

https://crates.io/crates/wasmtime-wasi-usb

JeffMboya (commented):
@Michielvk  Please mark as ready for review.

JeffMboya (approved):
LGTM

rodneyosodo (approved):
LGTM

PR #252, opened 2026-07-02
PROP-239 - Add latent tasks for on-demand FaaS execution

### LINKED ISSUE #239: Feature: Latent tasks to provide FaaS

Currently when one deploys a task, it immediately starts running exactly one copy of it, needing to listen on the network directly.  This probably isn't a long-term solution, since direct access the network is a possible attack vector, and different workloads might try to listen on the same ports.  In the long run, my prediction is that there will end up being some HTTP/MQTT proxy that routes incoming requests to where the task is located.

If this ends up happening, then a single long-running task will be a waste of memory when requests aren't coming in often, and a performance bottleneck when they are (since the component would need to be multithreaded and do a bunch of internal synchronisation to keep from corrupting its own memory).

It should be possible to deploy a task to a Proplet, so that it pulls the component and precompiles it, but doesn't immediately start executing.  Only when a request comes in should it actually instantiate the tasks's component, and concurrent requests can result in the component being instantiated multiple times and run in parallel.

Then, while waiting for requests, the task doesn't consume as much memory, and when many requests are coming in, they can more easily run in parallel as with traditional FaaS, and as well as being parallelised over multiple Proplets where a single worker doesn't have the capacity to manage all the requests on its own.

Nice-to-have

### Anything else?

_No response_

JeffMboya:
# What type of PR is this?

Feature: adds latent FaaS tasks that precompile on deploy and instantiate on demand.

## What does this do?

Deploys a task that precompiles its WASM component without running, then instantiates it per request in parallel.

Resolves #239

Yes. Proplet precompile/invoke and concurrent-invocation tests, plus a manager InvokeTask service test.

No README/API docs changed; usage is the new `tasks invoke <id>` CLI command.

### Notes

Latent flag persists only in memory/badger stores, matching existing Priority and Schedule handling.

### DIFF
--- a/manager/service.go
+++ b/manager/service.go
@@ -1462,40 +1462,26 @@ func (svc *service) updateResultsHandler(ctx context.Context, msg map[string]any
 	return nil
 }
 
-func (svc *service) updateInvokeResultsHandler(ctx context.Context, msg map[string]any) error {
-	taskID, ok := msg["task_id"].(string)
-	if !ok {
-		return errors.New("invalid task_id")
-	}
-	if taskID == "" {
-		return errors.New("task id is empty")
+func (svc *service) updateInvokeResultsHandler(_ context.Context, msg map[string]any) error {
+	invocationID, ok := msg["invocation_id"].(string)
+	if !ok || invocationID == "" {
+		return nil
 	}
 
 	results, _ := msg["results"].(string)
 	errMsg, _ := msg["error"].(string)
 
-	if invocationID, ok := msg["invocation_id"].(string); ok && invocationID != "" {
-		svc.pendingInvokesMu.Lock()
-		ch, waiting := svc.pendingInvokes[invocationID]
-		svc.pendingInvokesMu.Unlock()
-		if waiting {
-			select {
-			case ch <- invokeResult{results: results, err: errMsg}:
-			default:
-			}
+	svc.pendingInvokesMu.Lock()
+	ch, waiting := svc.pendingInvokes[invocationID]
+	svc.pendingInvokesMu.Unlock()
+	if waiting {
+		select {
+		case ch <- invokeResult{results: results, err: errMsg}:
+		default:
 		}
 	}
 
-	t, err := svc.GetTask(ctx, taskID)
-	if err != nil {
-		return err
-	}
-
-	t.Results = msg["results"]
-	t.Error = errMsg
-	t.UpdatedAt = time.Now()
-
-	return svc.taskRepo.Update(ctx, t)
+	return nil
 }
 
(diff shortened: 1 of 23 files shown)


proplet/src/service.rs
```diff
@@ -657,6 +666,43 @@ impl PropletService {
...
+                self.metrics.tasks_failed.inc();
+                self.metrics.tasks_running.dec();
+                self.publish_result(&task_id, Vec::new(), Some(e.to_string()))
+                    .await?;
+
+                return Err(e);
+            }
+
+            info!(
+                "Latent task {} precompiled and ready for on-demand invocation",
```
rodneyosodo, line 702: `tasks_running` is incremented at L539 for all start commands (including latent), but only decremented on the error path (L691). The success path just returns Ok(()) and leaks the counter.

proplet/src/service.rs
```diff
@@ -1043,6 +1089,83 @@ impl PropletService {
...
+        let task_id = req.id.clone();
+        let invocation_id = req.invocation_id.clone().unwrap_or_default();
+        let inputs = req.inputs.clone();
+        let env = req.env.unwrap_or_default();
+        let metrics = self.metrics.clone();
+
+        let invoke_span = tracing::info_span!("task.invoke.execute", task_id = %task_id);
+        tokio::spawn(
+            async move {
+                metrics.tasks_started.inc();
```
rodneyosodo, line 1127: `tasks_started` is already incremented during precompile (L538). This second increment on invoke means `tasks_started = 1 + N` while `tasks_completed = N` — the metric invariant is off by one.

proplet/src/service.rs
```diff
@@ -1043,6 +1089,83 @@ impl PropletService {
...
+                metrics.tasks_running.inc();
+
+                let result = runtime.invoke(task_id.clone(), inputs, env).await;
```
rodneyosodo, line 1130: If `runtime.invoke()` panics, `tasks_running.dec()` at L1129 is skipped. Consider wrapping in `catch_unwind` or noting the gap.

manager/service.go
```diff
@@ -1904,6 +1991,32 @@ func (svc *service) publishStop(ctx context.Context, t task.Task, propletID stri
...
+		p, err := svc.SelectProplet(ctx, t)
+		if err != nil {
+			return "", err
+		}
+
+		return p.ID, nil
+	}
+
+	return svc.taskPropletRepo.Get(ctx, t.ID)
+}
```
rodneyosodo, line 2010: `publishInvoke` uses `map[string]any` while every other publish method uses a typed struct. Also `InvokeRequest.env` exists but is never populated.

rodneyosodo (commented):
### Metrics observations while reviewing latent tasks

rodneyosodo (approved):
LGTM

PR #260, opened 2026-07-18
fix(compose): stop creating config.toml as a directory

### LINKED ISSUE #258: Bug: Provision wizard fails to create `docker/config.toml` file
### What were you trying to achieve?

Step 5.2 of the [Propeller K8s operator example](https://www.absmach.eu/docs/propeller/k8s/example/#52--run-the-provision-wizard)

### What are the expected results?

`docker/config.toml` file successfully created.

### What are the received results?

`Error: failed to create docker/config.toml file`

Actually, an empty **directory** `docker/config.toml/`  is created (not a plain file) for some reason, even when aborting the command from the beginning.

### Steps To Reproduce

```console
$ ./build/cli provision -f docker/config.toml
```

### In what environment did you encounter the issue?

Ubuntu 24.04 LTS

### Additional information you deem important

If I change the filename to something else like `docker/conf.toml`, the `conf.toml` file is created successfully, but the empty **directory** `docker/config.toml/`  mentioned earlier is still created! So my current workaround it to run the command with a `-f docker/conf.toml`, remove the `config.toml/` directory, and rename `conf.toml` to `config.toml`

floze-the-genius:
## What does this do?

Closes #258.

The provision wizard writes `docker/config.toml` as a regular file, but the Propeller Compose file used short bind-mount syntax for that path. If Compose is started before provisioning, Docker creates a missing bind source as a directory. The wizard then cannot replace that directory with the generated TOML file and reports `failed to create docker/config.toml file`.

This converts all three config mounts (manager, proplet, and proxy) to long bind syntax with `bind.create_host_path: false`.

Compose will now fail clearly when the provisioned config is missing instead of mutating the host path into a directory. Existing mount targets and runtime behavior are unchanged once the file exists.

Users who already have an empty `docker/config.toml/` directory still need to remove it once before running the provision command; Compose will not recreate it afterward.

## Verification

- Parsed `docker/compose.propeller.yaml` with Docker Compose v5.1.2.
- Inspected normalized Compose JSON and verified all 3 config mounts have `bind.create_host_path == false`.
- Ran `go test ./...` successfully.
- Ran `git diff --check` successfully.

### DIFF
--- a/docker/compose.propeller.yaml
+++ b/docker/compose.propeller.yaml
@@ -57,7 +57,11 @@ services:
       # MANAGER_MQTT_TLS_CLIENT_KEY: ${MANAGER_MQTT_TLS_CLIENT_KEY}
       # MANAGER_MQTT_TLS_INSECURE_SKIP_VERIFY: ${MANAGER_MQTT_TLS_INSECURE_SKIP_VERIFY}
     volumes:
-      - ./config.toml:/config.toml
+      - type: bind
+        source: ./config.toml
+        target: /config.toml
+        bind:
+          create_host_path: false
       # MQTT TLS certs (uncomment and change source paths when using ssl://):
       # - type: bind
       #   source: ./ssl/certs/ca.crt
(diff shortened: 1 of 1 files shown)


rodneyosodo:
Hello @floze-the-genius, thanks for your contribution, but we have a different approach as implemented by https://github.com/absmach/propeller/pull/272

PR #261, opened 2026-07-18
fix(ssl): create certificate output directory

### LINKED ISSUE #257: Bug: Missing directory `docker/ssl/certs`
### What were you trying to achieve?

In the directory of Propeller repository, run `make start-all`  (or `make ca` from `docker/ssl` folder).

### What are the expected results?

The command should succeed.

### What are the received results?

The command fails with the following error:

```
$ make start-all
...
req: Can't open "certs/ca.key" for writing, No such file or directory
```

### Steps To Reproduce

```
$ git clone  https://github.com/absmach/propeller.git
$ cd propeller
$ make start-all
```

### In what environment did you encounter the issue?

Ubuntu 24.04 LTS

### Additional information you deem important

Creating `docker/ssl/certs` manually fixes it. But for a long-term fix, this directory should be present in the repository or the script should create it if it does not exist.

floze-the-genius:
## What type of PR is this?

Bug fix.

## What does this do?

Closes #257.

A fresh clone does not contain `docker/ssl/certs`, so certificate targets fail when their first writer tries to create a file under `$(CRT_LOCATION)`.

This adds one idempotent `$(CRT_LOCATION)` directory target and makes every currently implemented certificate-producing target (`ca`, `server_cert`, and `client_cert`) use it as an order-only prerequisite. Existing directories are left untouched, and directory timestamps do not regenerate certificates.

The patch is rebased onto current `main`. Certificate-generation settings and the existing CA dependency of server/client certificates are unchanged.

## Verification

- Generated CA, server, and client certificates with real OpenSSL 3.6.3 from a missing nested output directory.
- Verified the generated server and client certificates against the generated CA.
- Confirmed a pre-existing output directory does not schedule another `mkdir`.
- `go test ./... -count=1`
- `go vet ./...`
- `golangci-lint v2.12.2 run --config ./.golangci.yaml` — 0 issues
- `git diff --check`

Documentation is unchanged because this restores the existing documented setup behavior.

### DIFF
--- a/docker/ssl/Makefile
+++ b/docker/ssl/Makefile
@@ -44,6 +44,11 @@ etc, etc.
 endef
 all: clean_certs ca server_cert client_cert coap_dtls_certs
 
+$(CRT_LOCATION):
+	mkdir -p $@
+
+ca server_cert client_cert: | $(CRT_LOCATION)
+
 # CA name and key is "ca".
 ca:
 	openssl req -newkey rsa:2048 -x509 -nodes -sha512 -days 1095 \


rodneyosodo:
@floze-the-genius rebase and implement this for all certs

floze-the-genius:
@rodneyosodo Rebased onto current `main` and applied the order-only `CRT_LOCATION` prerequisite to all three current certificate-producing targets: `ca`, `server_cert`, and `client_cert`. I also added focused Makefile integration coverage and verified the root flow with real OpenSSL; the exact commands are in the updated description. Ready for another look.

docker/ssl/makefile_test.go
```diff
@@ -0,0 +1,330 @@
+package ssl_test
```
rodneyosodo: No need for this

rodneyosodo (changes requested):

docker/ssl/makefile_test.go
floze-the-genius: Removed the Makefile integration test as requested. I also rebased the branch onto current `main`; the PR is now a single five-line `docker/ssl/Makefile` change. I revalidated the missing nested-directory flow with real OpenSSL, verified both generated certificates against the CA, and reran the Go, vet, lint, and diff gates. Ready for another look.

rodneyosodo (approved):

PR #277, opened 2026-08-18
Implement ELASTIC wasi-security

Michielvk:
# What type of PR is this?

New feature as part of the ELASTIC project. As discussed and agreed upon by all partners during the technical meeting in Belgrade.

## What does this do?

This PR allows a user to supply an additional config file to the Proplet. The Proplet parses this config file and configures Wasmtime accordingly.

Yes. All pass.

Yes. And an example config file is available.


rodneyosodo:
@Michielvk, can we use Metadata rather than having a new entry on the task?

Michielvk:
> @Michielvk, can we use Metadata rather than having a new entry on the task?

@rodneyosodo, this was the original plan, but it turns out that the Metadata field is only used by the Manager and stripped while creating the `startPayload` before sending it to the Proplet.

https://github.com/idlab-discover/propeller/blob/6d32fa570e9f40da0d499ebf0095be2f6e125549/manager/service.go#L2041-L2042

Since we need these additional configs in the Proplet to configure Wasmtime, we decided during the Belgrade meeting to add a new field instead that is intended to supply custom config to the Proplet directly.

Michielvk:
@rodneyosodo `Metadata` field solution tested and confirmed working locally with different policies.

proplet/README.md
```diff
@@ -149,3 +149,81 @@ url = "http://10.0.2.2:8082"
 Do not include a `file` field for encrypted workloads.
 
 For full TEE setup (KBS, image encryption, CVM provisioning), see the [Encrypted workloads guide](https://propeller.absmach.eu/docs/tee).
+
+## WASI security policy
```
rodneyosodo: Can we move this to rather than having it here https://github.com/absmach/propeller-docs

rodneyosodo (changes requested):
The merge is almost ready.

proplet/README.md
Michielvk: Yes, I will create a PR to the propeller-docs with examples as well.

rodneyosodo (approved):
LGTM

PR #287, opened 2026-09-02
Use shared keys and enable FluxMQ broker auth via Atom gRPC

rodneyosodo:
Signed-off-by: Rodney Osodo <socials@rodneyosodo.com>
# What type of PR is this?


arvindh123:
Review: request changes.

1. **P1 – BLOCKING: upgrade breaks existing MQTT clients.** `ATOM_BROKER_CREDENTIAL_KIND=shared_key` makes the native Atom callout accept only shared keys, but this PR creates them only for new `provision`, `add-proplets`, and `add-proxy` runs. Existing deployments have access tokens in `config.toml` (`MANAGER_API_KEY`, `PROPLET_API_KEY`, and `PROXY_API_KEY`), so they will all be rejected after the Compose upgrade. Please add a migration/rotation path that creates shared keys and updates deployed configuration before flipping the broker, or retain a compatibility path during rollout.

2. **P1 – BLOCKING if direct Docker Compose is supported; otherwise P2/non-blocking.** New Atom/FluxMQ certificate bind mounts are generated only through `make start-base` → `check-certs`. A clean checkout started with `docker compose ... up` has no such files and cannot start correctly. Please make preparation an explicit documented prerequisite or add a Compose-native initialization path, and test a clean-checkout start.

3. **P2 – NON-BLOCKING FUNCTIONALLY, but requires scope approval.** `docker/atom/atom-bootstrap.yaml` grants six `alarm_*` capabilities to every existing and future tenant admin. This is unrelated to the shared-key/native-callout migration and materially expands authorization. Please split it into a separate PR or document its Propeller dependency and test the intended boundary.

4. **P2 – NON-BLOCKING for an explicitly local development stack; BLOCKING for production or shared hosts.** `docker/ssl/Makefile` uses `chmod 0644` for Atom and FluxMQ private keys. Any local user can then impersonate the FluxMQ client to Atom. Please preserve restrictive modes and arrange appropriate container UID/GID access instead.

5. **Required coverage – NON-BLOCKING on its own, but needed to prove items 1 and 2.** Please add an integration test for FluxMQ↔Atom mTLS, MQTT shared-key authentication, publish/subscribe authorization, invalid-key denial, and an upgrade with pre-existing credentials.

arvindh123 (approved):
LGTM

PR #293, opened 2026-09-14
Adding complete support for `--wasi-pep` CLI arg / `metadata.elastic.wasi_pep` property

cdanger:
# What type of PR is this?

## What does this do?

As part of Elastic project, this completes the changes initiated in PR #277 (WASI security policy support) to add full support of the `--wasi-pep` CLI argument and `metadata.elastic.wasi_pep` property in Manager API for Task creation.
This enables the integration with not only THS "Lightweight ABAC component" (short name "WASI PEP") but also with any other `WasiSecurityProvider` (new interface) plugin that may create and provide a WASI security policy dynamically based on the configuration passed by `--wasi-pep` CLI arg / `metadata.elastic.wasi_pep` property, and other workload metadata.

The WASI Security Provider plugin (e.g. THS WASI PEP) is loaded as an external (shared) library (from `$PROPLET_PLUGIN_DIR/libWasiSecurityPlugin.so`), using the `dlopen2` crate, so that it may be implemented and maintained separately from Propeller. 

This feature is an alternative to the static WASI security policy passed via `--wasi-security` CLI arg / `metadata.elastic.wasi_security` property in Manager API. The Manager API and CLI have been modified to verify that `wasi_pep` and `wasi_security` are not used together at the same time.

A `lib` target has been added to proplet project's `Cargo.toml` to build a library containing the WASI Security structs, that WASI Security Provider plugins can then use as a dependency to implement the `WasiSecurityProvider` interface.

Relates to PR #277 .

Tests added in PR #277 .

There are explanatory comments in the code modifications. I can update Propeller's documentation as well, when the PR is merged.

### Notes

When the proplet returns an error because of bad wasi_pep / wasi_security input in a Start Task request (for example `proplet/src/service.rs`, line 524), the Manager seems to ignore the error and says the task has started successfully regardless. No error returned to the client. 
**Do you have some advice on how to return the error properly to the Manager? If it is possible?**


proplet/Cargo.toml
```diff
@@ -85,3 +88,9 @@ default = []
 
 [profile.release]
 strip = "symbols"
+
+# ELASTIC - The wasi_security structures are used by external WASI security providers, e.g. THS WASI PEP library.
+# TODO: make wasi_security.rs a separate library crate
```
rodneyosodo, line 93: Do we need this?

rodneyosodo (changes requested):

rodneyosodo (approved):
LGTM

